PR #30, opened 2015-02-06
The create Artifact request should have a non-empty version.

### LINKED ISSUE #29: The request should fail when creating an Artifact without a version
The artifact create request must have a non-null version value

maheshkelkar:
Added a cucumber test and check in Artifact class.
Removed the non-null restriction on version from DB representtion.

Fixes #29

### DIFF
--- a/features/api/artifact/creating.feature
+++ b/features/api/artifact/creating.feature
@@ -32,7 +32,7 @@ Feature: Artifact CREATE APIs
     """
 
   @error 
-  Scenario: The reuqest should fail when creating an Artifact without a version
+  Scenario: The request should fail when creating an Artifact without a version
 
     When I PUT to "/api/v1/artifacts" with:
     """
--- a/features/api/artifact/creating.feature
+++ b/features/api/artifact/creating.feature
@@ -31,17 +31,18 @@ Feature: Artifact CREATE APIs
       }
     """
 
-  @error @wip
-  Scenario: Creating an artifact without a version
+  @error 
+  Scenario: The reuqest should fail when creating an Artifact without a version
 
     When I PUT to "/api/v1/artifacts" with:
     """
       {
         "group" : "com.example.cucumber",
-        "name" : "cukes"
+        "name" : "cukes",
+        "sourceUrl" : "http://example.com/cucumber.jar"
       }
     """
-    Then the response should be 400
+    Then the response should be 422
 
   @wip @error
   Scenario: Creating an artifact with too large of a group or name
--- a/features/api/artifact/creating.feature
+++ b/features/api/artifact/creating.feature
@@ -32,6 +32,32 @@ Feature: Artifact CREATE APIs
     """
 
   @error 
+  Scenario: The request should fail when creating an Artifact without a group name
+
+    When I PUT to "/api/v1/artifacts" with:
+    """
+      {
+        "name" : "cukes",
+        "version" : "1.2.345",
+        "sourceUrl" : "http://example.com/cucumber.jar"
+      }
+    """
+    Then the response should be 422
+
+  @error 
+  Scenario: The request should fail when creating an Artifact without a name
+
+    When I PUT to "/api/v1/artifacts" with:
+    """
+      {
+        "group" : "com.example.cucumber",
+        "version" : "1.2.345",
+        "sourceUrl" : "http://example.com/cucumber.jar"
+      }
+    """
+    Then the response should be 422
+
+  @error 
   Scenario: The request should fail when creating an Artifact without a version
 
     When I PUT to "/api/v1/artifacts" with:
@@ -43,13 +69,6 @@ Feature: Artifact CREATE APIs
       }
     """
     Then the response should be 422
-
-  @wip @error
-  Scenario: Creating an artifact with too large of a group or name
-
-    Given an artifact with a name over 8192 characters
-    When I POST to "/api/v1/artifacts" with it
-    Then the response should be 400
 
   @error
   Scenario: The request should fail when creating an Artifact without a sourceUrl
@@ -63,3 +82,28 @@ Feature: Artifact CREATE APIs
       }
     """
     Then the response should be 422
+
+  @error 
+  Scenario: Creating an artifact with too large of a group name
+
+    When I PUT to "/api/v1/artifacts" with a group over 8192 characters
+    Then the response should be 422
+
+  @error 
+  Scenario: Creating an artifact with too large of a name
+
+    When I PUT to "/api/v1/artifacts" with a name over 8192 characters
+    Then the response should be 422
+
+  @error 
+  Scenario: Creating an artifact with too large of a version
+
+    When I PUT to "/api/v1/artifacts" with a version over 8192 characters
+    Then the response should be 422
+
+  @error 
+  Scenario: Creating an artifact with too large of a sourceUrl
+
+    When I PUT to "/api/v1/artifacts" with a sourceUrl over 8192 characters
+    Then the response should be 422
+
--- a/src/cucumber/groovy/step_definitions/HttpSteps.groovy
+++ b/src/cucumber/groovy/step_definitions/HttpSteps.groovy
@@ -3,7 +3,7 @@ import cucumber.api.*
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.ObjectMapper
 import org.joda.time.DateTime
-
+import java.util.Random 
 
 // Add functions to register hooks and steps to this script.
 this.metaClass.mixin(cucumber.api.groovy.Hooks)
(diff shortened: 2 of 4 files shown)


features/api/artifact/creating.feature
```diff
@@ -31,17 +31,18 @@ Feature: Artifact CREATE APIs
       }
     """
 
-  @error @wip
-  Scenario: Creating an artifact without a version
+  @error 
+  Scenario: The reuqest should fail when creating an Artifact without a version
```
rtyler: `request` typo

src/main/resources/db/migration/V2__artifacts_versions.sql
```diff
@@ -8,7 +8,7 @@ ALTER TABLE artifacts
         /*
          * 255 characters of version information should be enough for anybody!
          */
-        version VARCHAR(255) NOT NULL,
+        version VARCHAR(255),
```
rtyler: Why is this change here? there should be no changes to the database layer for this.

maheshkelkar: Trying to keep "version" & "sourceUrl" definitions consistent. I had added following comment in the commit description "Removed the non-null restriction on version from DB representtion.".

features/api/artifact/creating.feature
maheshkelkar: will fix and push again.

src/main/resources/db/migration/V2__artifacts_versions.sql
rtyler: This is a database constraint that needs to stay in place, a row with a NULL version is invalid

maheshkelkar: Why not same constraint on the sourceUrl? Yesterday we deemed it
unnecessary, so I was wondering why not make them consistent. I will back
it out.

rtyler: I could be misremembering, I think I said we shouldn't be doing request validation with database constraints . We should probably add a subsequent issue to update the database constraint for `sourceUrl`

maheshkelkar:
Apparently code review fixes for #29, somehow are coupled with of fix for #30.

The fix is ready for merge. Can somebody please take care of it?

src/cucumber/groovy/step_definitions/HttpSteps.groovy
```diff
@@ -25,11 +25,37 @@ When(~/^I PUT to "(.*?)" with:$/) { String path, String requestBody ->
...
+When(~/^I PUT to "(.*?)" with a (.*?) over ([1-9][0-9]*) characters$/) { String path, String var, int varSize ->
+
+    // Create a randomString of size varSize+100 
+    String alphabet = (('a'..'z') + ('A'..'Z')+('0'..'9')).join()
+    String randomString = ""
+    for ( i in 1..varSize+100 ) {
+        randomString += alphabet[ new Random().nextInt(alphabet.length()-1)]
+    }
+
+    print "MVK : $var\n"
```
rtyler: This print dshould not be here.

src/cucumber/groovy/step_definitions/HttpSteps.groovy
```diff
@@ -25,11 +25,37 @@ When(~/^I PUT to "(.*?)" with:$/) { String path, String requestBody ->
     response = putJsonToPath(path, requestBody)
 }
 
+When(~/^I PUT to "(.*?)" with a (.*?) over ([1-9][0-9]*) characters$/) { String path, String var, int varSize ->
```
rtyler: This is a *very* specialized step definition just for the artifact creation API which makes it a bit smelly to me.

What we're actually doing in this step is what shohuld be done primarily in the Given step, setting up the test state. I suggest refactoring this into:

```cucumber
Given I have an artifact
 And the group is over 8192 characters
When I PUT ...
Then the response should be 422
```

src/cucumber/groovy/step_definitions/HttpSteps.groovy
```diff
@@ -25,11 +25,37 @@ When(~/^I PUT to "(.*?)" with:$/) { String path, String requestBody ->
...
+
+    // Create a randomString of size varSize+100 
+    String alphabet = (('a'..'z') + ('A'..'Z')+('0'..'9')).join()
```
rtyler: This code is unnecessarily too fancy, just create a big ass string of 'a' characters or something :)

src/main/groovy/deploydb/models/Artifact.groovy
```diff
@@ -16,19 +16,26 @@ import org.hibernate.validator.constraints.NotEmpty
 @Table(name='artifacts')
 class Artifact extends AbstractModel {
 
+    @NotEmpty
+    @Size(min=1, max=8192)
```
rtyler: Does `@Size(min=1)` make the `@NotEmpty` annotation redundant now?

maheshkelkar: > This print should not be here.
[mvk] Done. Removed

> This is a *very* specialized step definition just for the artifact
creation API which makes it a bit smelly to me.

> What we're actually doing in this step is what should be done primarily
in the Given step, setting up the test state. I suggest refactoring this
into:

> Given I have an artifact
>  And the group is over 8192 characters
> When I PUT ...
> Then the response should be 422

[mvk] The goal is to create one of the attributes with string of 8192+
chars. "Given I have an artifact" creates an artifact, we dont want to
do that. We want to to tweak the requestBody string for an attribute.

[mvk] I assume the parser, parses one step at a time, hence I tried to
compress entire statement in one.

> This code is unnecessarily too fancy, just create a big ass string of 'a' characters or something :)

[mvk] ok. done

Does @Size(min=1) make the @NotEmpty annotation redundant now?

[mvk] no. I tried. Apparently @Size only kicks in if object exists.

rtyler:
@maheshkelkar responding to code comments inline in email doesn't work out very well, better to use the web interface for it :/

rtyler:
I should also mention, that this pull request has an incorrect title which makes it quite confusing

src/cucumber/groovy/step_definitions/HttpSteps.groovy
maheshkelkar: We want to create one of the attributes (group, name, version, or sourceUrl) in the requestBody with a string of 8192+ chars. 

"Given I have an artifact" creates an artifact. So I assum we dont want to do that. We want to to tweak the requestBody string for an attribute. 

I assume the parser, parses one step at a time, hence I tried to compress entire statement in one. 

How about the following?
E.g. 
When I PUT to "/api/v1/artifacts" 
  And the group is over 8192 characters

maheshkelkar:
@rtyler Pull request for #29 and #33 somehow got merged under this. Perhaps my comment in #33 commit was "Fixes #33" and I think it should have been "Fix #33". Is there a way to fix this?

src/cucumber/groovy/step_definitions/HttpSteps.groovy
rtyler: @maheshkelkar you're right, "Given I have an artifact" is not what we want since it's putting things in the DB. Let's add a new Given step like: "Given I have an artifact request" e.g.:

```cucumber
Given I have an artifact request
And the group is over 8192 characters
When I PUT ..
```

I think it's pretty important to keep the "setup" in the Given stanza

maheshkelkar: *For cucumber:*

```
  @error
  Scenario: Creating an artifact with too large of a group name 2

    Given I have an artifact request
    And the group is over 8192 characters    
    When I PUT to "/api/v1/artifacts"
    Then the response should be 422
```
*Steps*:
```
Given(~/^I have an artifact request$/) { ->
}

And(~/^the (.*?) is over ([1-9][0-9]*) characters$/) { String var, int varSize ->

    // Create a randomString of size varSize+1
    String randomString = "test-".padRight(varSize+1, "a")

    group = var == "group" ? randomString : "com.example.cucumber"
    name  = var == "name" ? randomString : "cukes"
    version = var == "version" ? randomString : "1.2.345"
    sourceUrl = var == "sourceUrl" ? randomString : "http://example.com/cucumber.jar"
    
    requestBody = """ {
        "group" : "$group",
        "name" : "$name",
        "version" : "$version",
        "sourceUrl" : "$sourceUrl"
      }
    """    
}    

When(~/^I PUT to "(.*?)"$/) { String path ->
    response = putJsonToPath(path, requestBody)
}
```

src/cucumber/groovy/step_definitions/HttpSteps.groovy
maheshkelkar: done
```
    String randomString = "test-".padRight(varSize+1, "a")
```

rtyler:
I'm going to merge this and @maheshkelkar is going to promise to submit better pull requests with far less git branch confusion in the future

PR #34, opened 2015-02-06
This is intended to communicate ideas and not as an actual code change a...

tommysullivan:
...t this time

Main ideas:

1. Use Artifactory REST API directly and do not create an abstraction around it at this time?
2. Use POST to /collection-like/ URLs to create new objects and expect location header with URL of newly created resource?
3. Use PUT to /specific/id URLs to replace resource that already has a known URL?
4. Use ETAG for caching of resources and to avoid write collisions?
5. Use content type negotiation and vendor specific media types for resource representations for flexibility and discoverability reasons?
6. Use URI to refer to related concepts such as expressing that a deployment resource is related to an environment resource rather than a simple string?


rtyler:
I'm going to try to respond to the main ideas here, instead of line-by-line in the feature file changes (which I agree are a good way to communicate things, @polynomial is smoking drug cigarettes).

> Use Artifactory REST API directly and do not create an abstraction around it at this time?

DeployDB does need some unique sense of an Artifact stored in the database in order to answer the questions "Where is X now?" and "Where has X been?"

I'm also partial to storing the unique (group, name, version) for an Artifact in the database as it's also more resilient to repository changes.

> Use POST to /collection-like/ URLs to create new objects and expect location header with URL of newly created resource?

I think returning a `Location` header of the newly created resource is a good idea. I don't want to do that instead of the JSON response.though. I think any API client should receive a proper `Artifact` object back from what they've created

> Use PUT to /specific/id URLs to replace resource that already has a known URL?

I don't think that 'replacing' should actually occur in any cases. There's some PATCH methods which have been scoped out for Deployment, but that's not replacing the resource so much as modifying in place. I don't think the DeployDB worldview should allow for replacing data whole-hog


> Use ETAG for caching of resources and to avoid write collisions?

I'm not sure I understand the underlying concern here, HTTP is not a great protocol to attempt to avoid write collisions with, first-past-the-post is always going to win with the writes to the database.

> Use content type negotiation and vendor specific media types for resource representations for flexibility and discoverability reasons?

I really don't understand this ask. Not that I'm necessarily saying "`application/json` should be good enough for everybody", but what do we have to gain from the additional complexity of special content types all over the place?


> Use URI to refer to related concepts such as expressing that a deployment resource is related to an environment resource rather than a simple string?

I'm liking this idea, but I need to ponder more on it and how it might manifest it self in APIs. Internal to DeployDB, a Deployment must be referred to its ID, but I understand how it can be useful to pass around URIs to clients. I'm concerned with binding an API-looking URI into the datamodel and reducing flexibility to change things later on.



To give you an idea of where my primary inspiration comes from as far as REST API design, I'm quite partial to the [GitHub v3 APIs](https://developer.github.com/v3/) and generally try to emulate them as much as possible.

Issue #32, opened 2015-02-06
Fix the request artifact with support for pageNumber=0 and pageSize=0


GiriDandu:
This test case is failing because hibernate has a bug in the setMaxResult(0) returning all the records instead of single record. See this: https://hibernate.atlassian.net/browse/HHH-8989

```
 Given there are artifacts                                                                                  # ArtifactSteps.groovy:15
    When I GET "/api/v1/artifacts/by-module/com.example.cucumber:cucumber-artifact?pageNumber=0&perPageSize=0" # HttpSteps.groovy:16
    Then the response should be 200                                                                            # HttpSteps.groovy:33
    And the body should be JSON:                                                                               # HttpSteps.groovy:41
      """
        [{
          "id" : 2,
          "createdAt" : "{{created_timestamp}}",
          "group" : "com.example.cucumber",
          "name" : "cucumber-artifact",
          "version" : "1.0.2",
          "sourceUrl" : "http://example.com/maven/com.example.cucumber/cucumber-artifact/1.0.2/cucumber-artifact-1.0.2.jar"
        }]
      """
      Assertion failed: 
      
      assert bodyNode == expectedNode
             |        |  |
             |        |  [{"id":2,"createdAt":"2015-02-06T22:27:26.977Z","group":"com.example.cucumber","name":"cucumber-artifact","version":"1.0.2","sourceUrl":"http://example.com/maven/com.example.cucumber/cucumber-artifact/1.0.2/cucumber-artifact-1.0.2.jar"}]
             |        false
             [{"id":1,"createdAt":"2015-02-06T22:27:26.977Z","group":"com.example.cucumber","name":"cucumber-artifact","version":"1.0.1","sourceUrl":"http://example.com/maven/com.example.cucumber/cucumber-artifact/1.0.1/cucumber-artifact-1.0.1.jar"},{"id":2,"createdAt":"2015-02-06T22:27:26.977Z","group":"com.example.cucumber","name":"cucumber-artifact","version":"1.0.2","sourceUrl":"http://example.com/maven/com.example.cucumber/cucumber-artifact/1.0.2/cucumber-artifact-1.0.2.jar"}]
 ````

rtyler:
In the example aobve, the `pageSize` is zero, why shouldn't that just return an empty list then?

GiriDandu:
[...]

PR #38, opened 2015-02-10
Issue 10 expose list of artificats

GiriDandu:
Add testing based on multiple artifacts based on review comments

### DIFF
--- a/features/api/artifact/reading.feature
+++ b/features/api/artifact/reading.feature
@@ -51,41 +51,44 @@ Feature: Artifact READ APIs
         "sourceUrl" : "http://example.com/maven/com.example.cucumber/cucumber-artifact/1.0.1/cucumber-artifact-1.0.1.jar"
       }]
     """
-@freezetime
-Scenario: Fetching an artifact by the (group, name) tuple, pageNumber and perPageSize
+  @freezetime
+  Scenario: Fetching an artifact by the (group, name) tuple, pageNumber and perPageSize
 
-    Given there are artifacts
-    When I GET "/api/v1/artifacts/by-module/com.example.cucumber:cucumber-artifact?pageNumber=0&perPageSize=5"
-    Then the response should be 200
-    And the body should be JSON:
-    """
-      [{
-        "id" : 2,
-        "createdAt" : "{{created_timestamp}}",
-        "group" : "com.example.cucumber",
-        "name" : "cucumber-artifact",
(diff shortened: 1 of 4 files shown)


features/api/artifact/reading.feature
```diff
@@ -51,38 +51,66 @@ Feature: Artifact READ APIs
...
     When I GET "/api/v1/artifacts/by-module/com.example.cucumber:cucumber-artifact?pageNumber=1&perPageSize=5"
     Then the response should be 404
 
+  @error
   Scenario: Fetching an artifact by the (group, name) tuple with invalid data type for pageSize
 
     When I GET "/api/v1/artifacts/by-module/com.example.cucumber:cucumber-artifact?pageNumber=-1&perPageSize=0xdeadbeef"
     Then the response should be 400
 
+@wip
```
rtyler: Based on some discussion we had previously (can't find the link now) I was under the impression that this API with a zero page size and a zero per-page, we'd return an empty list"

src/test/groovy/deploydb/models/ArtifactSpec.groovy
```diff
@@ -3,6 +3,16 @@ package deploydb.models
 import spock.lang.*
 
 class ArtifactSpec extends Specification {
+
+    private Artifact artifact
+
+    def setup() {
```
rtyler: Why is this here? It's not being used by anything

src/test/groovy/deploydb/models/ArtifactSpec.groovy
```diff
@@ -14,6 +24,15 @@ class ArtifactSpec extends Specification {
 
 class ArtifactWithArgsSpec extends Specification {
 
+    private Artifact artifact
+
+    def setup() {
+        artifact = new Artifact('spock.group',
```
rtyler: Same comment as above, I don't understand why this setup() method is here

features/api/artifact/reading.feature
GiriDandu: Agreed, Thats why I move this test to @wip and I am tracking this with another issue #32.

src/test/groovy/deploydb/models/ArtifactSpec.groovy
GiriDandu: Put back the change that @maheshkelkar made.  Fix with latest commit

PR #42, opened 2015-02-11
Convert YAML files into Service Models/objects

### LINKED ISSUE #36: We need some ability to deserialize YAML into Models
We've got [some examples}(http://hackers.lookout.com/deploydb/internals.html#_services) where we need to define something that approximates a "model" in the DeployDB codebase.

Dropwizard itself has support for a single configuration file, but we need some ability to define YAML-backed models.

For example, `services/auditlog.yml` would eventually be translated into a `deploydb.models.Service` instance for auditlog.

I don't want this issue to include the handling of loading those yaml files necessarily, but the code to translate configuration-driven models from YAML into objects.

maheshkelkar:
Added Service object model. Added tests to read various YAML files
and generate Service objects.

Fixes #36


src/main/groovy/deploydb/models/Service.groovy
```diff
@@ -0,0 +1,24 @@
+package deploydb.models
+
+import com.google.common.collect.Lists
+import com.fasterxml.jackson.annotation.JsonProperty
+import javax.validation.constraints.NotNull
+import org.hibernate.validator.constraints.NotEmpty
+
+class Service {
```
rtyler: This class should have some javadocs associated with it and its members.

src/test/groovy/deploydb/models/ServiceSpec.groovy
```diff
@@ -0,0 +1,88 @@
...
+class ServiceWithArgsSpec extends Specification {
+
+    private final Validator validator = Validation.buildDefaultValidatorFactory().getValidator()
+    private final ConfigurationFactory<Service> factory =
+            new ConfigurationFactory<>(Service.class, validator, Jackson.newObjectMapper(), "dw")
+
+    def "Loading of valid service config file suceeds"() {
+        given:
+        ClassLoader classLoader = getClass().getClassLoader()
+        File validFile = new File(classLoader.getResource("services/test-valid.yml").toURI())
```
rtyler: Why not refactor these two lines into a helper method in the class which returns a `File` for a given file? It's used in enough specifications

src/test/resources/services/test-invalid.yml
```diff
@@ -0,0 +1 @@
+name: Boop
```
rtyler: boop

src/main/groovy/deploydb/models/Service.groovy
```diff
@@ -0,0 +1,24 @@
...
+
+    @NotNull
+    @JsonProperty
+    String name
+
+    @NotEmpty
+    @JsonProperty
+    List<String> artifacts
```
rtyler: Can you add additional validation to ensure that the list of artifacts follows the `group:name` convention?

rtyler:
Overall a good looking first pass!

src/main/groovy/deploydb/models/Service.groovy
maheshkelkar: what do you mean by javadocs? comments or comments with tags?

src/main/groovy/deploydb/models/Service.groovy
maheshkelkar: I have added max size of 10 for number of artifacts, pipelines and promotions

maheshkelkar:
Took care of the code review comments #36

PR #51, opened 2015-02-18
Expose Services via read API

### LINKED ISSUE #43: Expose Services via a read API
Since Services are configuration driven, I think all we need is an API to list all services (e.g. `/services`) and an API to list details about a specific service (e.g. `/services/<name>`)

maheshkelkar:
Added a ModelRegistry to manage config driven data objects (e.g. Service)
in the memory. Added an REST API to return all or a specific by name
Service instance. Added cucumber and spock unit tests.

Fixes #43

### DIFF
--- a/src/cucumber/groovy/deploydb/cucumber/AppHelper.groovy
+++ b/src/cucumber/groovy/deploydb/cucumber/AppHelper.groovy
@@ -18,6 +18,9 @@ import org.hibernate.SessionFactory
 import org.hibernate.Transaction
 import org.hibernate.context.internal.ManagedSessionContext
 
+import deploydb.resources.ModelRegistry
+import deploydb.models.Service
+
 class AppHelper {
     private StubAppRunner runner = null
     private Client jerseyClient = null
@@ -44,6 +47,15 @@ class AppHelper {
         }
     }
 
+    /**
+     *  Execute the {@link Closure} with a proper Service Registry
+     *
+     * @param c (required) Closure to execute 
+     */
+    void withServiceRegistry(Closure c) {
+        ModelRegistry<Service> serviceRegistry = this.runner.serviceRegistry
+        c.call(serviceRegistry)
+    }
 
     String processTemplate(String buffer, Map scope) {
         DefaultMustacheFactory mf = new DefaultMustacheFactory()
--- a/src/cucumber/groovy/deploydb/cucumber/AppHelper.groovy
+++ b/src/cucumber/groovy/deploydb/cucumber/AppHelper.groovy
@@ -18,7 +18,7 @@ import org.hibernate.SessionFactory
 import org.hibernate.Transaction
 import org.hibernate.context.internal.ManagedSessionContext
 
-import deploydb.resources.ModelRegistry
+import deploydb.registry.ModelRegistry
 import deploydb.models.Service
 
 class AppHelper {
@@ -53,8 +53,7 @@ class AppHelper {
      * @param c (required) Closure to execute 
      */
     void withServiceRegistry(Closure c) {
-        ModelRegistry<Service> serviceRegistry = this.runner.serviceRegistry
-        c.call(serviceRegistry)
+        c.call(this.runner.serviceRegistry)
     }
 
     String processTemplate(String buffer, Map scope) {
@@ -71,7 +70,7 @@ class AppHelper {
         if (this.jerseyClient == null) {
             ClientConfig clientConfig = new ClientConfig()
             clientConfig.connectorProvider(new ApacheConnectorProvider())
-            this.jerseyClient = ClientBuilder.newClient(clientConfig);
+            this.jerseyClient = ClientBuilder.newClient(clientConfig)
         }
         return this.jerseyClient
     }
--- a/src/main/groovy/deploydb/resources/ModelRegistry.groovy
+++ b/src/main/groovy/deploydb/resources/ModelRegistry.groovy
@@ -1,0 +1,81 @@
+package deploydb.resources
+
+import java.io.File
+import java.util.*
+import java.io.IOException
+import javax.validation.Validation
+import javax.validation.Validator
+import io.dropwizard.jackson.Jackson
+import com.google.common.io.Resources;
+import io.dropwizard.configuration.ConfigurationFactory
+import com.fasterxml.jackson.annotation.JsonProperty
+import io.dropwizard.configuration.ConfigurationException
+import io.dropwizard.configuration.ConfigurationParsingException
+import io.dropwizard.configuration.ConfigurationValidationException
+import java.util.concurrent.ConcurrentHashMap
+import java.lang.IllegalArgumentException
+
+/**
+ * Model Registry object
+ *
+ * It creates the configuration driven model objects. The key for Model 
+ * instances is String. Objects are stored in hashMap for later retrieval.
+ */
+class ModelRegistry<T> {
+    private final ConcurrentHashMap<String, T> modelTable
+    private final Validator validator
+    private final ConfigurationFactory<T> factory
+
+   /**
+    * Creates a new Model Registry for the given class.
+    *
+    * @param klass          the Model class
+    */
+    public ModelRegistry(Class<T> klass) {
+        this.validator = Validation.buildDefaultValidatorFactory().getValidator()
+        this.factory = new ConfigurationFactory<>(klass,
+                                                  validator,
+                                                  Jackson.newObjectMapper(), "")
+        this.modelTable = new ConcurrentHashMap<String, T>()
+    }
+
+    /**
+     * Insert an extant model in the table
+     *
+     * @param model          the Model instance
+     */
+    void insert(T model) throws IllegalArgumentException {
+        //if (model.name == null) throw IllegalArgumentException
+        modelTable.put(model.name, model)
+    }
+    
+    /**
+     * Instantiate a new Model object by parsing contents of the input file (T) 
+     *
+     * @param File name, file is present in the resources directory
+     */
+    T createFromFile(String filename)
+        throws Exception, ConfigurationParsingException,
+               ConfigurationValidationException {
+        File inputFile = new File(Resources.getResource(filename).toURI())
+        T model = factory.build(inputFile)
+        this.insert(model)
+        return model
+    }
+
+    /**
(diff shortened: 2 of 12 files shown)


src/cucumber/groovy/deploydb/cucumber/AppHelper.groovy
```diff
@@ -44,6 +47,15 @@ class AppHelper {
         }
     }
 
+    /**
+     *  Execute the {@link Closure} with a proper Service Registry
+     *
+     * @param c (required) Closure to execute 
+     */
+    void withServiceRegistry(Closure c) {
+        ModelRegistry<Service> serviceRegistry = this.runner.serviceRegistry
```
rtyler: Why bother with an intermediate variable here? Just pass `this.runnerServiceRegistry` into the `Closure.call` method

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
+package deploydb.resources
```
rtyler: I think this should be in the `deploydb.registry` package (wihch doesn't exist _yet_)

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
...
+
+import java.io.File
+import java.util.*
```
rtyler: IMO wildcard imports are a code smell. That said, Groovy might already make everything in `java.util` visible in the default scoping

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
...
+import java.io.IOException
+import javax.validation.Validation
+import javax.validation.Validator
+import io.dropwizard.jackson.Jackson
+import com.google.common.io.Resources;
```
rtyler: Unnecessary semicolon

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
...
+import io.dropwizard.configuration.ConfigurationFactory
+import com.fasterxml.jackson.annotation.JsonProperty
+import io.dropwizard.configuration.ConfigurationException
+import io.dropwizard.configuration.ConfigurationParsingException
+import io.dropwizard.configuration.ConfigurationValidationException
+import java.util.concurrent.ConcurrentHashMap
+import java.lang.IllegalArgumentException
```
rtyler: I believe Groovy includes everything in `java.lang` in the default scope making this import unnecessary

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
+package deploydb.resources
+
+import java.io.File
```
rtyler: Will you please order these impors alphabetically

src/cucumber/groovy/deploydb/cucumber/AppHelper.groovy
maheshkelkar: Fixed. leftover effect.

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
...
+    private final ConcurrentHashMap<String, T> modelTable
+    private final Validator validator
+    private final ConfigurationFactory<T> factory
+
+   /**
+    * Creates a new Model Registry for the given class.
+    *
+    * @param klass          the Model class
+    */
+    public ModelRegistry(Class<T> klass) {
```
rtyler: Since this is a generic class, I'm not sure that we actually need to pass this `Class. object into the constructor, can you refer to `T` inside the `ConfigurationFactory` instantiation below?

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
...
+import java.lang.IllegalArgumentException
+
+/**
+ * Model Registry object
+ *
+ * It creates the configuration driven model objects. The key for Model 
+ * instances is String. Objects are stored in hashMap for later retrieval.
+ */
+class ModelRegistry<T> {
+    private final ConcurrentHashMap<String, T> modelTable
```
rtyler: I'm not sure why this is marked `final`, we're certainly going to be modifying it later

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
...
+import java.util.concurrent.ConcurrentHashMap
+import java.lang.IllegalArgumentException
+
+/**
+ * Model Registry object
+ *
+ * It creates the configuration driven model objects. The key for Model 
+ * instances is String. Objects are stored in hashMap for later retrieval.
+ */
+class ModelRegistry<T> {
```
rtyler: Please add:

```groovy
import groovy.transform.TypeChecked
```

So this class can be annotated with `@TypeChecked`, internal classes which have strong APIs like this one I've been trying to make sure we enable type checking unless we have a good reason not to

src/main/groovy/deploydb/resources/ModelRegistry.groovy
rtyler: `public` is implicit in Groovy, you don't need to add it to anything

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
...
+        T model = factory.build(inputFile)
+        this.insert(model)
+        return model
+    }
+
+    /**
+     * Return list of all model objects
+     */
+    List<T> getAllModels() {
+        return new ArrayList<T>(modelTable.values())
```
rtyler: why isn't this just returning `modelTable.values()`?

src/main/groovy/deploydb/resources/ServiceResource.groovy
```diff
@@ -0,0 +1,69 @@
...
+
+import com.codahale.metrics.annotation.Timed
+import com.google.common.base.Optional
+import io.dropwizard.jersey.caching.CacheControl
+import io.dropwizard.jersey.params.*
+import io.dropwizard.hibernate.UnitOfWork
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+import javax.validation.Valid
```
rtyler: Valid isn't used anywhere here, it should be removed

src/main/groovy/deploydb/resources/ModelRegistry.groovy
maheshkelkar: Since ModelRegistry<T> instantiates another template ConfigureFactory<T>, the innertmost instantiation does not like T.class and fails

src/main/groovy/deploydb/resources/ServiceResource.groovy
```diff
@@ -0,0 +1,69 @@
...
+        this.serviceRegistry = serviceRegistry
+    }
+
+    /**
+     * Returns all Service objects
+     */
+    @GET
+    @UnitOfWork
+    @Timed(name = "get-requests")
+    public List<Service> getAll(@Context HttpHeaders headers) {
```
rtyler: Headers are not being used here, the argument should be removed

src/main/groovy/deploydb/resources/ServiceResource.groovy
```diff
@@ -0,0 +1,69 @@
+package deploydb.resources
+
+import com.codahale.metrics.annotation.Timed
+import com.google.common.base.Optional
```
rtyler: I don't believe Optional is being used so this import can be removed

src/main/groovy/deploydb/resources/ModelRegistry.groovy
maheshkelkar: Hit exception, failed to convert modelTable.values() to list()

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
...
+                                                  Jackson.newObjectMapper(), "")
+        this.modelTable = new ConcurrentHashMap<String, T>()
+    }
+
+    /**
+     * Insert an extant model in the table
+     *
+     * @param model          the Model instance
+     */
+    void insert(T model) throws IllegalArgumentException {
```
rtyler: IMO this should be `put` to mirror the `AbstractMap` API

src/main/groovy/deploydb/resources/ModelRegistry.groovy
```diff
@@ -0,0 +1,81 @@
...
+    List<T> getAllModels() {
+        return new ArrayList<T>(modelTable.values())
+    }
+
+    /**
+     * Locate a Model by string name
+     *
+     * @param name The model instance's name (e.g. "fun-as-a-service")
+     */
+    T getModelByName(String name) {
```
rtyler: IMO this should be `get` to mirror the `AbstractMap` API.

That allows for:

```groovy
Service stupidService = registry['stupidserv']
```

src/main/groovy/deploydb/resources/ModelRegistry.groovy
rtyler: I think it is safe for the type to be `Collection<T>`, look at the javadocs, Collection is a superinterface of List

src/main/groovy/deploydb/resources/ModelRegistry.groovy
rtyler: Were you passing `T` or `T.class`, I think the first one is supposed to work

src/main/groovy/deploydb/resources/ModelRegistry.groovy
maheshkelkar: @rtyler do I put this under separate directory? or it will accept deploydb.registry as is.

rtyler: Yep, separate directory

PR #52, opened 2015-02-18
Initial cut at restructuring of APIs

rtyler:
See the commit messages for more specific details

### DIFF
--- a/features/api/artifact/reading.feature
+++ b/features/api/artifact/reading.feature
@@ -22,6 +22,19 @@ Feature: Artifact READ APIs
       }
     """
 
+  @wip
+  Scenario: Fetching an artifact with the proper media type
+
+    It is encouraged to specify the version of the API you would like to
+    receive inside of a Accept header. If this header is not specified than the
+    client is expected to receive the v1 version of the JSON API
+
+    Given there is an artifact
+    When I GET "/api/artifacts/1" with custom headers:
+      | Header Name | Value                            |
+      | Accept      | application/vnd.deploydb.v1+json |
+    Then the response should be 200
+
   Scenario: Fetching an artifact by ID that doesn't exist
 
     When I GET "/api/artifacts/1"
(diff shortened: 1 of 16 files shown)


features/api/artifact/reading.feature
```diff
@@ -22,15 +22,28 @@ Feature: Artifact READ APIs
...
+
+    It is encouraged to specify the version of the API you would like to
+    receive inside of a Accept header. If this header is not specified than the
+    client is expected to receive the v1 version of the JSON API
+
+    Given there is an artifact
+    When I GET "/api/artifacts/1" with custom headers:
+      | Header Name | Value                            |
+      | Accept      | application/vnd.deploydb.v1+json |
+    Then the response should be 200
```
GiriDandu: Can we add a test error case for invalid media type, to make sure we return 406

rtyler: Totally, technically that's what this scenario does right now whether I want it to or not though  :(

src/cucumber/groovy/step_definitions/ArtifactSteps.groovy
```diff
@@ -21,3 +21,25 @@ Given(~/^there are artifacts$/) { ->
...
+    name  = var == "name" ? randomString : "cukes"
+    version = var == "version" ? randomString : "1.2.345"
+    sourceUrl = var == "sourceUrl" ? randomString : "http://example.com/cucumber.jar"
+
+    requestBody = """ {
+        "group" : "$group",
+        "name" : "$name",
+        "version" : "$version",
+        "sourceUrl" : "$sourceUrl"
+      }
```
GiriDandu: Not sure why this is moved here.

rtyler: While I was updating some steps I noticed that this was in the wrong file (HttpSteps) which was not the right place for this to live

PR #54, opened 2015-02-19
Issue 18 webhook invocation during deployment

GiriDandu:


features/api/webhook/deploymentCreated.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Webhook invocation  when deployment is created
+
+  @wip @webhook
+  Scenario: Webhooks should be invoked when artifacts are created
+    Given the webhooks configuration:
+    """
+      deployment:
+        created: "http://{{webhook_url}}"
+    """
+    And the service configuration:
```
rtyler: I don't think that including this statement (and the configuration) is actualy useful for the documentation. I suggest hiding it inside of a step called something like:

```cucumber
Given a service is configured
```

features/api/webhook/deploymentCreated.feature
```diff
@@ -0,0 +1,42 @@
...
+      - com.github.lookout.puppet:puppet-foas
+      - com.github.lookout:puppet-mysql
+      - com.example.cucumber:cucumber-artifact
+    pipelines:
+      - devtoprod
+    promotions:
+      - status-check
+      - jenkins-smoke
+    """
+    When I PUT to "/api/v1/artifacts" with an artifact
```
rtyler: POST

features/api/webhook/promotionCompletedFail.feature
```diff
@@ -0,0 +1,41 @@
...
+        completed: "http://{{webhook_url}}"
+    """
+    And there is an artifact
+    When I trigger deployment PATCH "/api/deployments/1" with:
+    """
+      {
+        "promotion":{
+          "name": "status-check"
+          "url" : "http://local.lookout.com/jenkins/job-id/2/results"
+          "status" : "FAIL"
```
rtyler: This is not valid JSON, there's a missing trailing brace.

I think the API should be a POST to `/api/deployments/1/promotions` as well

src/main/groovy/deploydb/models/Webhook/Deployment.groovy
```diff
@@ -0,0 +1,37 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import org.hibernate.validator.constraints.NotEmpty
+
+import javax.validation.constraints.Size
+
+/**
+ *
+ */
+class WebhookDeployment {
```
rtyler: Before this PR can be merged I'd like to see these models moved into the `deploydb.models.webhooks` package, e.g.

```groovy
package deploydb.models.webhooks

class Deployment {
}
```

src/main/groovy/deploydb/models/Webhook/Deployment.groovy
```diff
@@ -0,0 +1,37 @@
+package deploydb.models
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import org.hibernate.validator.constraints.NotEmpty
+
+import javax.validation.constraints.Size
+
+/**
+ *
```
rtyler: Please fill in the javadocs before this is merged on what these different models map to

src/main/groovy/deploydb/models/Webhook/Deployment.groovy
```diff
@@ -0,0 +1,37 @@
...
+ */
+class WebhookDeployment {
+
+    @Size(max=2)
```
rtyler: Why is there any max size at all?

src/main/groovy/deploydb/models/Webhook/Webhook.groovy
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Class to store webhook configuration
+ */
+class Webhook {
+
+    /**
+     * List of Artifact names (max 10).
+     */
+    @NotEmpty
+    @Size(max=10)
```
rtyler: No need for a size restriction, or a notempty validator

tommysullivan:
Coming from a event notification handler and triggerer, here are our cukes. We can adjust but need to do so soon cuz apparently we need this to actually work in a real environment including kicking off tests that do not exist yet in less than 8 business days.

https://gist.github.com/tommysullivan/208f30c8e4748457d777

features/api/webhook/deploymentCompleted.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Webhook invocation  when deployment is completed
+
+  @wip @webhook
+  Scenario: Webhooks should be invoked when deployment is completed
+    Given the webhooks configuration:
+    """
+      deployment:
+        completed: "http://{{webhook_url}}"
+    """
+    And there is an artifact
```
tommysullivan: i like to see the config in the cukes because it serves as documentation at a glance without needing to go into code or search for matching regexp and all that. i think this should be the default approach.

features/api/webhook/deploymentCompleted.feature
```diff
@@ -0,0 +1,33 @@
...
+    When I trigger deployment PATCH "/api/deployments/1" with:
+    """
+      {
+        "status" : "COMPLETED"
+      }
+    """
+    Then the webhook should be invoked with the JSON:
+    """
+      {
+        "id" : 1,
```
tommysullivan: in HATEOAS the URL of the deployment resource should be provided so that clients may know of it without having to have a priori knowledge of "how to hook a 1 to the rest of the string that would comprise the url". if you must have the "1" as its own distinct property (i wonder if there is a legit use case yet?) then fine, simply add the fully qualified href to another property. Make sense?

features/api/webhook/deploymentCompleted.feature
```diff
@@ -0,0 +1,33 @@
...
+        "artifact" : {
```
tommysullivan: For now we will expect single artifact per POST... or can there be an array?

features/api/webhook/deploymentCompleted.feature
```diff
@@ -0,0 +1,33 @@
...
+          "id" : 1,
+          "group" : "com.example.cucumber",
+          "name" : "cucumber-artifact",
```
tommysullivan: what "name" should we expect for bluffdale deploy? what names should we expect for other service / component deploys that should cause bluffdale integration tests to run (such as kafka, storm, metron, etc).

features/api/webhook/deploymentCompleted.feature
```diff
@@ -0,0 +1,33 @@
...
+          "version" : "1.0.1",
+          "sourceUrl" : "http://example.com/maven/com.example.cucumber/cucumber-artifact/1.0.1/cucumber-artifact-1.0.1.jar",
+          "createdAt" : "{{created_timestamp}}"
+        },
+        "service" : "fun as service"
+        "environment" : "dev-apha",
```
tommysullivan: should we expect this to be stage1? (current MTP testing stage environment?)

ben is planning to spin up a new environment dynamically this sprint. presumably this will have some other name. given that we receive this name, how do we use it to construct the full url we will need? ideally this process of resolving the url will be encapsulated by deploydb. if it uses consul, fine. if it uses something else, fine. let us know. short term workarounds are totally cool with us.

features/api/webhook/deploymentCreated.feature
```diff
@@ -0,0 +1,42 @@
...
+    Then the webhook should be invoked with the JSON:
```
tommysullivan: for v0.1 this is cool and maybe even beyond; but we would like to advance for your consideration that the design be to host deployment resources yourself, then you simply POST an event that a given deployment resource has changed. the payload would simply be the URL of the existing deployment resource you are already hosting. That way the client can, upon receipt of the notification, use content negotiation to retrieve a representation of the recently changed deployment resource using a format that they understand, so that different versions of clients may co-exist simultaneously. This also makes it easier for proxies or related systems to pass around "a web reference" to the recently changed deployment without passing this entire body. just a later version thing to consider.

features/api/webhook/deploymentCompleted.feature
tommysullivan: also it might be a symptom. you are POSTing to web hook a deployment representation. The HTTP semantics would imply this means you are creating a deployment.

really what you are "CREATING" is a notification that a deployment (on your system) has been updated. So the representation really is just a description of the update, and a link to the resource that was updated - your deployment. If we did a GET on that URL then it would make sense to only have the 1 as the ID because the URL is already known to the requesting client.

features/api/webhook/deploymentCompleted.feature
rtyler: :+1: 

There's some cukes that still need to be written around Service, Promotion,  Environment and Pipeline configurations FWIW

features/api/webhook/deploymentCompleted.feature
rtyler: We don't have any concept of a base url *yet*, when we get to that point, I tihnk we'll want to add an `url` and/or `html_url` at some point in the future

features/api/webhook/deploymentCompleted.feature
rtyler: It's not well exposed in the documentation yet, but there exists a one-to-one relationship between Deployment and Artifact

features/api/webhook/deploymentCompleted.feature
rtyler: `name` in this context is only the name of the artifact itself. I won't reference internal artifact names since this is a public forum, but we're following the Java-style artifact convention of group and name.

features/api/webhook/deploymentCompleted.feature
rtyler: The name of the environment listed here would be whatever is configured for DeployDB. In our cases something like "stage1" would be the conventional name of an environment and therefore configured in DeployDB as such

Issue #53, opened 2015-02-18
Use yaml filename as a key (id) for Service Model instance


maheshkelkar:
All the configuration driven models (i.e. service, environment, pipeline, webhooks, promotions, etc) are created from yaml file. Use the filename as the "id" in the object instance created from the config.

rtyler:
In this change will you also change the property name in the configuration files from "name" to "description" per our IRL discussion yesterday?

rtyler:
Isn't this fixed now? You can set up commit messages to auto-close issues by the way with `Fixes #53` at the bottom of the commit message

maheshkelkar:
Yeah. I had a #53 in each commit. Not sure why it didnt close. But in git :)

PR #55, opened 2015-02-19
Added variable, ident, in the Service model; renamed name to description

maheshkelkar:
Added missing ServiceResourcesSpec tests. Moved load models from yaml to
seperate class ModelLoadear.

References #53

### DIFF
--- a/src/main/groovy/deploydb/ModelLoader.groovy
+++ b/src/main/groovy/deploydb/ModelLoader.groovy
@@ -13,10 +13,9 @@ import javax.validation.Validation
 import javax.validation.Validator
 
 /**
- * Model Registry object
+ * Model Loader object
  *
- * It creates the configuration driven model objects. The key for Model 
- * instances is String. 
+ * It instantiates a model object from the yaml configuration.
  */
 @TypeChecked
 class ModelLoader<T> {
--- a/src/main/groovy/deploydb/ModelLoader.groovy
+++ b/src/main/groovy/deploydb/ModelLoader.groovy
@@ -1,0 +1,60 @@
+package deploydb
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import com.google.common.io.Resources
+import groovy.transform.TypeChecked
+import io.dropwizard.configuration.ConfigurationFactory
+import io.dropwizard.jackson.Jackson
+import io.dropwizard.configuration.ConfigurationException
+import io.dropwizard.configuration.ConfigurationParsingException
+import io.dropwizard.configuration.ConfigurationValidationException
+import java.io.File
+import javax.validation.Validation
+import javax.validation.Validator
+
+/**
+ * Model Registry object
+ *
+ * It creates the configuration driven model objects. The key for Model 
+ * instances is String. 
+ */
+@TypeChecked
+class ModelLoader<T> {
+    private final Validator validator
+    private final ConfigurationFactory<T> factory
+
+   /**
+    * Creates a new Model Registry for the given class.
+    *
+    * @param klass          the Model class
+    */
+    ModelLoader(Class<T> klass) {
+        this.validator = Validation.buildDefaultValidatorFactory().getValidator()
+        this.factory = new ConfigurationFactory<>(klass,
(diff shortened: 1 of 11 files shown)


src/main/groovy/deploydb/ModelLoader.groovy
```diff
@@ -0,0 +1,60 @@
...
+import io.dropwizard.jackson.Jackson
+import io.dropwizard.configuration.ConfigurationException
+import io.dropwizard.configuration.ConfigurationParsingException
+import io.dropwizard.configuration.ConfigurationValidationException
+import java.io.File
+import javax.validation.Validation
+import javax.validation.Validator
+
+/**
+ * Model Registry object
```
rtyler: This javadoc is now out of date

src/main/groovy/deploydb/ModelLoader.groovy
```diff
@@ -0,0 +1,60 @@
...
+class ModelLoader<T> {
+    private final Validator validator
+    private final ConfigurationFactory<T> factory
+
+   /**
+    * Creates a new Model Registry for the given class.
+    *
+    * @param klass          the Model class
+    */
+    ModelLoader(Class<T> klass) {
```
rtyler: This constructor I don't believe is valid anymore since this is a generic class

src/main/groovy/deploydb/ModelLoader.groovy
```diff
@@ -0,0 +1,60 @@
...
+        return model
+    }
+
+    /**
+     * Get identifier for model from the the input file 
+     *
+     * @param File name, file is present in the resources directory
+     */
+    String getIdent(String filename) {
+        return filename.substring(filename.lastIndexOf('/') + 1,
```
rtyler: Instead of doing this manual string manipulation, why not save the `File` in the constructor to a member variable and then use `File.getName()`

src/main/groovy/deploydb/models/Service.groovy
```diff
@@ -11,12 +11,21 @@ import org.hibernate.validator.constraints.NotEmpty
...
+     * registry
+     */
+    @JsonProperty
+    String ident
+
+    /**
+     * Service description
      */
     @NotEmpty
     @Size(max=8192)
```
rtyler: This size restriction doesn't actually apply I don't think

src/main/groovy/deploydb/ModelLoader.groovy
maheshkelkar: It still needs it. Because we are instantiating template within template.

src/main/groovy/deploydb/ModelLoader.groovy
maheshkelkar: ModelLoader is used to read multiple files and create objects (e.g. service). I dont think we can save the file.

src/main/groovy/deploydb/models/Service.groovy
maheshkelkar: Since we are reading description from the yaml file. I think it will apply. It wont apply on "ident"

maheshkelkar:
travis failed, because concurrentHashMap keeps flipping order of the services and it messes the script

maheshkelkar:
@rtyler @GiriDandu this pull request is ready for merge. I assume Giri's pull request is not merged yet

src/main/groovy/deploydb/models/Service.groovy
rtyler: There's no size limit on a String though, that descripton we're not going to persist to a database at any point

maheshkelkar: I think it does. E.g. if description in yaml file is a string with size longer than 8192, then yaml parsing will fail

src/main/groovy/deploydb/ModelLoader.groovy
rtyler: That's a fair point, I'm going to ponder some more on this, I think we can refactor this API a bit more as our configuration-model needs

PR #56, opened 2015-02-20
Fix the build error.

### LINKED ISSUE #18: Web hooks should be invoked when a Deployment is created
Unclear as of yet what the contents of the webhook payload might be or how these webhooks might be created. Either way, Deployment creation should enqueue N webhooks to execute

rtyler:
-Fix #18


rtyler:
This commit message isn't very useful, but I'll go ahead and merge it. It gives the reader zero context to what the actual change is (which "Moving loadFromString into ModelTable where it belongs" might)

PR #58, opened 2015-02-23
Issue 18 webhook invocation during deployment

GiriDandu:
Add spock tests for webhook


src/test/groovy/deploydb/models/webhooks/WebhookSpec.groovy
```diff
@@ -0,0 +1,83 @@
...
+}
+
+class WebhookWithArgsSpec extends Specification {
+
+    private final ModelLoader<Webhook> webhookModelLoader =
+            new ModelLoader<Webhook>(Webhook.class)
+
+    def "Loading of valid webhook config file succeeds"() {
+        given:
+        Webhook webhook = webhookModelLoader.load('webhooks/test-valid.yml')
```
maheshkelkar: I think we should remove these files and define the file here as a string (Just like cucumber test). And load it via loadFromString() function. That way we can avoid creating multiple/duplicate files for each model.

src/main/groovy/deploydb/models/Webhook/Webhook.groovy
```diff
@@ -14,13 +14,27 @@ class Webhook {
...
     @JsonProperty
-    List<Promotion> webhookPromotionList
+    Promotion promotion
+
+    /**
+     *  Access function for deployment
+     */
+    public Deployment getDeployment(){
+        return this.deployment
+    }
```
maheshkelkar: we dont need this, because groovy generates getter/setter functions

src/main/groovy/deploydb/models/Webhook/Webhook.groovy
```diff
@@ -14,13 +14,27 @@ class Webhook {
...
+
+    /**
+     *  Access function for promotion
+     */
+    public Promotion getPromotion(){
+        return this.promotion
```
maheshkelkar: same as above. note needed

src/main/groovy/deploydb/models/Webhook/Promotion.groovy
```diff
@@ -15,7 +15,14 @@ class Promotion {
...
     @JsonProperty
-    List<String> completed
+    private List<String> completed
+
+    /**
+     * Access function for completed
+     */
+    public List<String> getCompleted(){
+        return this.completed
+    }
```
maheshkelkar: I think by defining private, you are hiding the completed object here. But, in one of my committs, @rtyler suggested keeping it all public.

src/main/groovy/deploydb/models/Webhook/Deployment.groovy
```diff
@@ -14,31 +14,60 @@ class Deployment {
      * List of URI associated with deployment started state
      */
     @JsonProperty
-    List<String> started
+    private List<String> started
```
maheshkelkar: I think by defining private, you are hiding the completed object here. But, in one of my committs, @rtyler suggested keeping it all public.

src/main/groovy/deploydb/models/Webhook/Deployment.groovy
```diff
@@ -14,31 +14,60 @@ class Deployment {
...
+     * Access function for verified
+     */
+    public List<String> getVerified(){
+        return this.verified
+    }
 
     /**
      * Empty constructor used by Jackson for object deserialization
      */
     Deployment() {}
```
maheshkelkar: We have 2 Deployment objects; 1> depolydb.models.Deployment & 2> deploydb.models.webhooks.Deployment. It may create some confusion.

src/test/groovy/deploydb/models/webhooks/WebhookSpec.groovy
```diff
@@ -0,0 +1,83 @@
...
+
+        then:
+        thrown(ConfigurationParsingException)
+    }
+
+    def "Loading a valid webhook with configuration inline succeeds"(){
+        given:
+        Webhook webhook = webhookModelLoader.loadFromString("deployment:\n" +
+                "  created:\n" +
+                "    - http://jenkins.example.com/job/notify-deployment-created/build")
```
maheshkelkar: Since "created" is a list. Can we add a test to have multiple urls?

src/test/resources/webhooks/test-emptymodels.yml
```diff
@@ -0,0 +1,10 @@
+deployment:
```
maheshkelkar: we can get rid of this file and move this content to string inside the spock/cucumber tests.

src/test/groovy/deploydb/models/webhooks/WebhookSpec.groovy
GiriDandu: Agreed, next time in this directory we will remove all the files and make them inline

src/main/groovy/deploydb/models/Webhook/Deployment.groovy
GiriDandu: I am not a big fan of all the member variables in a class public and have auto generated get/set functions. But to make it consistent with rest of the deploydb code, I will follow the suggested approach and move the members public

src/main/groovy/deploydb/models/Webhook/Deployment.groovy
GiriDandu: I think directory path should help us with this :-)

src/test/groovy/deploydb/models/webhooks/WebhookSpec.groovy
GiriDandu: Agreed, I will add this

src/test/resources/webhooks/test-emptymodels.yml
GiriDandu: Sure, when we remove all the test-**.yml files,  we will make this change

src/main/groovy/deploydb/ModelLoader.groovy
```diff
@@ -56,8 +59,12 @@ class ModelLoader<T> {
     T loadFromString(String content)
             throws Exception, ConfigurationParsingException,
                     ConfigurationValidationException {
-        File.createTempFile('tmp', '.yml').write(content)
-        return load('tmp.yml')
+        File tmpFile = File.createTempFile('tmp', '.yml', new File('./build/tmp/'))
```
rtyler: `./build/tmp` is not a gauranteed file path to be available. If I'm running `deploydb` in a daemon mode, which will happen in the future, then I believe this will fail.

If this funtion is only ever meant to be invoked from a test context then it should be noted as such and eventually moved elsewhere

src/main/groovy/deploydb/models/Webhook/Deployment.groovy
```diff
@@ -14,31 +14,60 @@ class Deployment {
...
      * List of URI associated with deployment verified state
      */
     @JsonProperty
-    List<String> verified
+    private List<String> verified
+
+    /*
+     * Access function for created
+     */
+    public List<String> getCreated(){
```
rtyler: In Groovy `public` is the default visibility, so these `public` declarations are unnecessary, you can safely remove them

src/main/groovy/deploydb/models/Webhook/Deployment.groovy
rtyler: Agreed with @GiriDandu namespacing is important :)

src/main/groovy/deploydb/models/Webhook/Webhook.groovy
rtyler: :+1:

src/main/groovy/deploydb/ModelLoader.groovy
GiriDandu: Yes, this function is used only for test. I will update the comment. And we will eventually move it from there

Issue #59, opened 2015-02-23
Expose Environments via REST APIs


maheshkelkar:
Add GET by name or GET all environments

PR #60, opened 2015-02-23
Define REST apis for Environment model

maheshkelkar:
References #59

### DIFF
--- a/features/api/environment/reading.feature
+++ b/features/api/environment/reading.feature
@@ -3,10 +3,11 @@ Feature: Environment READ APIs
   As a RESTful client or user
   I should be able to read information about environments
 
-  @wip
-  Scenario: Fetching all environments 
 
-    Given a integ environment configuration:
+  @wip
+  Scenario: Fetching all environments
+
+    Given an environment configuration named "integ":
     """
     description: "DeployDB Primary Integration"
     webhooks:
@@ -16,16 +17,6 @@ Feature: Environment READ APIs
         completed:
           - http://jenkins.example.com/job/integ-deploy-completed/build
     """
-    And another dev-alpha environment configuration:
-    """
-    description: "DeployDB Primary dev-alpha"
-    webhooks:
-      deployment:
-        created:
-          - http://jenkins.example.com/job/dev-alpha-deploy-created/build
-        completed:
-          - http://jenkins.example.com/job/dev-alpha-deploy-completed/build
-    """
     When I GET "/api/environments"
     Then the response should be 200
     And the body should be JSON:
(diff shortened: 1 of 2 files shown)


features/api/environment/reading.feature
```diff
@@ -0,0 +1,98 @@
+Feature: Environment READ APIs
+
+  As a RESTful client or user
+  I should be able to read information about environments
+
+  @wip
+  Scenario: Fetching all environments 
+
+    Given a integ environment configuration:
```
rtyler: "integ" means nothing as far as DeployDB is concerned, I suggest changing this to:

```cucumber
Given an environment configuration:
```

features/api/environment/reading.feature
```diff
@@ -0,0 +1,98 @@
...
+    """
+    description: "DeployDB Primary Integration"
+    webhooks:
+      deployment:
+        created:
+          - http://jenkins.example.com/job/integ-deploy-created/build
+        completed:
+          - http://jenkins.example.com/job/integ-deploy-completed/build
+    """
+    And another dev-alpha environment configuration:
```
rtyler: I don't think this "and" step is necessuary, you should remove it

features/api/environment/reading.feature
```diff
@@ -0,0 +1,98 @@
...
+    webhooks:
+      deployment:
+        created:
+          - http://jenkins.example.com/job/dev-alpha-deploy-created/build
+        completed:
+          - http://jenkins.example.com/job/dev-alpha-deploy-completed/build
+    """
+    When I GET "/api/environments"
+    Then the response should be 200
+    And the body should be JSON:
```
rtyler: :+1:

features/api/environment/reading.feature
```diff
@@ -0,0 +1,98 @@
...
+                "started" : [],
+                "created" : ["http://jenkins.example.com/job/dev-alpha-deploy-created/build"],
+                "completed" : ["http://jenkins.example.com/job/dev-alpha-deploy-completed/build"]
+            },
+            "promotions" : {
+                "completed" : []
+            }
+        }
+      }]
+    """
```
rtyler: There should be two new lines separating scenarios

features/api/environment/reading.feature
```diff
@@ -0,0 +1,98 @@
...
+                "started" : [],
+                "created" : ["http://jenkins.example.com/job/integ-deploy-created/build"],
+                "completed" : ["http://jenkins.example.com/job/integ-deploy-completed/build"]
+            },
+            "promotions" : {
+                "completed" : []
+            }
+        }
+      }
+    """
```
rtyler: 2 * \n

features/api/environment/reading.feature
maheshkelkar: Without that "And", we need to manipulate strings in order to split and save into 2 seperate files.

rtyler: I'm suggesting only putting one environment in the response below, I don't know what value we get from having both

maheshkelkar: We are testing getAll() functionality; also "And" is not necessary. I can repeat the "Given" twice.

rtyler: What I'm suggesting is that it's not something that's useful from a documentation standpoint. Sure in the Spock tests we should verify this much more, but an Array respoinse with one element I think is fine for the cukes

maheshkelkar:
@rtyler @GiriDandu I am incorporated all the comments. I think its good to go.

rtyler:
Woohoo! :+1:

PR #63, opened 2015-02-24
Defined environment model, backend for REST APIs, added spock tests

maheshkelkar:
Instead of using separate files for vairous yaml test files, used the extant loadFromString feature in the ModelLoader - better readability and documentation. Ditto for cucumber tests.

References #8


src/main/groovy/deploydb/DeployDBApp.groovy
```diff
@@ -92,22 +94,41 @@ class DeployDBApp extends Application<DeployDBConfiguration> {
...
+         * Instantiate DAO objects
+         */
         final ArtifactDAO adao = new ArtifactDAO(hibernate.sessionFactory)
         final DeploymentDAO ddao = new DeploymentDAO(hibernate.sessionFactory)
+
+        /**
+         * Instantiate registries for in memory storage
+         */
         serviceRegistry = new ModelRegistry<Service>()
+        environmentRegistry = new ModelRegistry<deploydb.models.Environment>()
```
rtyler: FYI, the `deploydb` part of the namespacing is unnecessary since this class is already in the `deploydb` package

src/main/groovy/deploydb/models/Environment.groovy
```diff
@@ -0,0 +1,80 @@
...
+     * registry
+     */
+    @JsonProperty
+    String ident
+
+    /**
+     * Environment description
+     */
+    @NotEmpty
+    @Size(max=8192)
```
rtyler: There's no size restriction on description

src/test/groovy/deploydb/models/EnvironmentSpec.groovy
```diff
@@ -0,0 +1,91 @@
...
+
+class EnvironmentWithArgsSpec extends Specification {
+    private final ModelRegistry<Environment> environmentRegistry =
+            new ModelRegistry<>()
+    private final ModelLoader<Environment> environmentLoader =
+            new ModelLoader<>(Environment.class)
+
+    def "Loading of valid environment config file suceeds"() {
+        given:
+        String fileContents =
```
rtyler: This is unnecessarily hard to read, Groovy supports multiline strings which would make this test far more readable and maintainable IMO:

```
➜  ~  groovysh 
x = """Groovy Shell (2.3.9, JVM: 1.7.0_65)
Type ':help' or ':h' for help.
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
groovy:000> x = """this 
groovy:001> is a multiline
groovy:002> string"""
===> this 
is a multiline
string
groovy:000> println x
this 
is a multiline
string
===> null
groovy:000> 

```

src/test/groovy/deploydb/models/EnvironmentSpec.groovy
```diff
@@ -0,0 +1,91 @@
...
+        String fileContents = "j&&&&"
+        Environment environment = environmentLoader.loadFromString(fileContents)
+
+        then:
+        thrown(ConfigurationParsingException)
+    }
+
+    def "Loading a invalid environment config file throws throws a validation exception"() {
+        when:
+        String fileContents =
```
rtyler: Multiline string this too plz

src/test/groovy/deploydb/models/ServiceSpec.groovy
```diff
@@ -20,14 +20,25 @@ class ServiceSpec extends Specification {
...
+            new ModelRegistry<>()
     private final ModelLoader<Service> serviceLoader =
-            new ModelLoader<Service>(Service.class)
+            new ModelLoader<>(Service.class)
 
     def "Loading of valid service config file suceeds"() {
         given:
-        Service service = serviceLoader.load('services/test-valid.yml')
-        service.ident = serviceLoader.getIdent('services/test-valid.yml')
+        String fileContents =
```
rtyler: m-m-m-m-multiline w00tw00t

rtyler:
Mostly looks good, some minor changes that I'd like you to make before merging though please

src/main/groovy/deploydb/DeployDBApp.groovy
maheshkelkar: It conflicts with io.dropwizard.setup.Environment and caused my tests to fail

src/test/groovy/deploydb/models/EnvironmentSpec.groovy
maheshkelkar: Yes. But yaml parser does not like indentations. And if I skip indentations e.g.
    String fileContents = '''
description: foo 
info: bar
    ""

It messes up the code.

src/main/groovy/deploydb/DeployDBApp.groovy
rtyler: I get that, I'm saying you can refer to it as `models.Environment` and things like that in the future, you don't necessarily need to use the fully qualified package name

src/test/groovy/deploydb/models/EnvironmentSpec.groovy
maheshkelkar: Sorry I meant to say:
Parser doesnt like (Extra indents):

```
    String fileContents = '''
        description: foo 
        info: bar
    '''
```
And following doesnt look good in code:
```
    String fileContents = '''
description: foo 
info: bar
'''
```

rtyler: The latter is fine IMO, a bunch of manual string concatenations is definitely more error prone

maheshkelkar: BTW this is not manual. IntelliJ is doing this for us. You sure you want me to do it like this. We may have to make similar change for spocks for all models.

src/main/groovy/deploydb/DeployDBApp.groovy
maheshkelkar: I tried doing this; but following combinations failed to compile:

```
import deploydb.models.environment
private ModelRegistry<models.Environment> environmentRegistry = new ModelRegistry<>()
```
And:
```
import deploydb.models
private ModelRegistry<models.Environment> environmentRegistry = new ModelRegistry<>()
```
Only this worked. But a> we are importing too much, b> then every usage of Environment has to be referred as models.Envinronment
```
import deploydb.*
private ModelRegistry<models.Environment> environmentRegistry = new ModelRegistry<>()
```

Issue #9, opened 2014-12-30
Allow defining a Pipeline via YAML


rtyler:
Pipeline:

> a set of deployments and promotions defining the relationships between each environment

Current thinking of `pipelines/cd-prod-pipeline1.yml`

```yaml
environments:
  - dev-alpha
  - dev-beta
  - integ
  - preprod
  - prod
```

I'm currently thinking about pipelines as a plural because one might have a couple different continuous delivery pipelines going on at once. For example, web services might be deployed with a simple dev -> prod pipeline, but less stable or more problematic services would go through different UAT environments before production.

### Open Questions

 * Should we be able to specify additional promotion criteria in the pipeline yaml file? E.g. promotion to the `prod` environment always requires a clean status check from `preprod` *and* a manual promotion?
 * Are environments defined anywhere else other than tacitly in the pipeline file?

rtyler:
I've further defined this in [this document](http://hackers.lookout.com/deploydb/internals.html#_pipelines) which roughly approximates a specification

maheshkelkar:
@GiriDandu can we close thie issue? I think its merged

PR #62, opened 2015-02-23
Add read API for promotions

GiriDandu:
Fix issue #61. Add read API for promotions

PR #65, opened 2015-02-24
Add RESTful API for Pipeline read feature.

GiriDandu:
Fix Issue #9


rtyler:
I'm merging this but we *must* change these APIs after 0.1.0, exposing the pipeline data model as nothing more than a representation of what is configured on disk I don't think is terrifically useful.

For example, the "environment" in a response should really be a full `Environment` model, which is going to necessitate #62 being merged and some refactoring going on.

PR #67, opened 2015-02-24
Added codenarc

rkuris:
'gradle check' is how you run it, producing an HTML report

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,6 +8,7 @@ plugins {
 apply plugin: 'eclipse'
 apply plugin: 'groovy'
 apply plugin: 'application'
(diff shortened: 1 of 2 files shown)


rtyler:
Where will this put an HTML report? If it goes into `docs/html5` we can publish it on the site too

rkuris:
build/reports/codenarc/cucumber.html is the default location; I think it can be changed but I'd have to look into it.

rtyler:
@rkuris nah that's fine, that's where test reports and things like that go already

rtyler:
This work has been incorporated into #69

Issue #75, opened 2015-02-25
Update Deployment REST APIs as discussed in DeployDb WorkFlow


maheshkelkar:
GET /api/deployments - returns all deployments
GET /api/deployments/{id} - returns deployment for {id}
PATCH /api/deployments/{id} w/ state=created/completed/failed 
PATCH /api/deployments/{id}/promotions w/ promotion identifier, status & results URL

Issue #19, opened 2015-01-22
Allow reporting of Promotion Criteria succes via an HTTP API


rtyler:
Related to #3 somewhat

For integration test runs required to satisfy Promotion Criteria, a test runner will need to be able to hit a RESTful API to report the status of the integration test.

The API should require:

 * environment
 * artifact version
 * test suite name
 * test suite result
 * more info URL
 * ???

rtyler:
At least in 0.1 we need to define this API, we don't necessarily need to implement it though.

rtyler:
@GiriDandu I think you may have partially started defining this with the Promotion Result stuff we discussed around  #54

PR #76, opened 2015-02-25
Refactored the deployment REST API as per the WorkFlow doc

maheshkelkar:
References #75


features/api/deployment/reading.feature
```diff
@@ -22,13 +58,266 @@ Feature: Deployment READ APIs
...
           "createdAt" : "{{created_timestamp}}"
         },
-        "environment" : "pre-production",
-        "status" : "STARTED",
+        "environment" : "pre-prod",
+        "service" : "faas",
+        "state" : "STARTED",
+        "promotionStatus" : [{
+          "ident" : "jenkins-smoke",
+          "state" : "inProgress",
```
rtyler: Since state is likely to map to an `Enum` internally, I would put this as `IN_PROGRESS` instead

features/api/deployment/reading.feature
```diff
@@ -22,13 +58,266 @@ Feature: Deployment READ APIs
...
+          "resultsUrl : ""
+         },
+         {
+          "ident" : "jenkins-smoke",
+          "state" : "success",
```
rtyler: `SUCCESS`

features/api/deployment/reading.feature
```diff
@@ -22,13 +58,266 @@ Feature: Deployment READ APIs
...
+          "resultsUrl : ""
+        }]
+        "createdAt" : "{{created_timestamp}}"
+      }
+    """
+
+  @freezetime @wip
+  Scenario: Fetching an deployment by pageNumber=0 and perPageSize=0 returns all deployments
```
rtyler: This is incorrect, @GiriDandu and I had gone through this previously. Requesting something like this should return an empty list. Returning "all deployments" is an unbounded set which has a high likelihood of breaking things horrible

features/api/deployment/updating.feature
```diff
@@ -44,3 +66,47 @@ Feature: Deployment UPDATE APIs
       }
     """
     Then the response should be 404
+
+
+  @wip
+  Scenario: Updating the Promotion status for a deployment
```
rtyler: This is in the wrong file, it should be something separate from the `deployments` feature

features/api/deployment/updating.feature
```diff
@@ -44,3 +66,47 @@ Feature: Deployment UPDATE APIs
...
+
+    Given there is a deployment
+    When I PUT "/api/deployments/1/promotions" with:
+    """
+      {
+        "name"  : "jenkins-smoke",
+        "status" : "SUCCESS",
+        "resultsUrl" : "http://local.lookout.com/jenkins/job-id/2/results"
```
rtyler: I think we should name this `infoUrl` it's not necessarily the results directly in all cases, but where one would find more information about the promotion

features/api/deployment/updating.feature
maheshkelkar: ok. I will move it to features/api/promotionResults/update.feature As a part of #19

PR #78, opened 2015-02-25
Issue 66 model pipeline

GiriDandu:
*The model pipeline will change post 0.1.* We will keep this issue open

Add code to load pipeline configuration
Add spock/cucumber tests for pipeline model


gradle.properties
```diff
@@ -1,3 +1,3 @@
-org.gradle.daemon=true
+org.gradle.daemon=false
```
maheshkelkar: Are we sure we want to commit this? This allowed us to run groovysh

GiriDandu: I think it will be useful for now for quick tests. If we think compile is taking long, then we can change the flag to true

features/api/pipeline/reading.feature
```diff
@@ -3,77 +3,115 @@ Feature: Pipeline READ APIs
...
-      - dev-beta:
-      - integ:
-      - preprod:
-          promotions:
-             prod-preflight
-             manual
-      - prod
+         dev-alpha:
+           promotions:
+              -
```
rtyler: What's with the empty list?

features/api/pipeline/reading.feature
```diff
@@ -3,77 +3,115 @@ Feature: Pipeline READ APIs
...
-             "preprod" : [
-                "promotions": ['prod-preflight', 'manual']
-             ]
-             "prod" : []
-         ]
+          "ident" : "devtoprod",
+          "description": "Development to production pipeline",
+          "environments": {
+            "dev-alpha": {
+              "promotions": [""]
```
rtyler: A single empty string array shouldn't be valid

src/main/groovy/deploydb/models/Pipeline/Environment.groovy
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    int hashCode() {
+        return Objects.hash(this.promotions)
+    }
+
+    /**
+     * Stringfy the environment
+     */
+    String toString(){
+        String output
```
rtyler: Why not just:

```groovy
return "promotions: ${promotions}"
```

instead of this temporary output

src/main/groovy/deploydb/models/Pipeline/Pipeline.groovy
```diff
@@ -0,0 +1,82 @@
+package deploydb.models.Pipeline
```
rtyler: Not sure why this is in a separate package? Additionally packages and directory names should be all lowered cased

src/main/groovy/deploydb/models/Pipeline/Pipeline.groovy
```diff
@@ -0,0 +1,82 @@
...
+     * Description of the pipeline
+     */
+    @JsonProperty
+    String description
+
+    /**
+     * list of environments that are part of this pipeline
+     */
+    @JsonProperty
+    Map<String, Environment> environments
```
rtyler: This property doesn't make sense to me, we don't want to embed an Environment definition inside of the pipeline definition, we just want to refer to the `ident` for a given Environment

src/main/groovy/deploydb/models/Pipeline/Pipeline.groovy
```diff
@@ -0,0 +1,82 @@
...
+     */
+    @Override
+    int hashCode() {
+        return Objects.hash(this.ident, this.description, this.environments)
+    }
+    /**
+     * Stringfy the pipeline
+     */
+    String toString(){
+        String output
```
rtyler: The temporary variable is unnecessary here as well

src/main/groovy/deploydb/resources/PipelineResource.groovy
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Return all the configured pipelines
+     */
+    @GET
+    @UnitOfWork
+    @Timed(name = "get-requests")
+    List<Pipeline> getAll(){
+        List<Pipeline> pipelineList = this.pipelineRegistry.getAll()
+
+        if(pipelineList.isEmpty()){
```
rtyler: There should really be spacing here (and elsewhere in this file), e.g.

```groovy
if (pipelineList.isEmpty()) {
```

Readability is important!

src/test/groovy/deploydb/models/pipelines/PipelineSpec.groovy
```diff
@@ -0,0 +1,85 @@
+package deploydb.models.Pipeline
```
rtyler: Lower-case package name

features/api/pipeline/reading.feature
GiriDandu: I modified this reading feature, i see that you commented on updated diff below

src/main/groovy/deploydb/models/Pipeline/Pipeline.groovy
GiriDandu: Will update the directory names

src/main/groovy/deploydb/models/Pipeline/Environment.groovy
GiriDandu: ok, i will update to return them

src/main/groovy/deploydb/models/Pipeline/Pipeline.groovy
GiriDandu: I agree, but in yml file for pipeline we are defining "promotions" under "evniroments" and we can't nest "promotions" without making into a map.

src/main/groovy/deploydb/models/Pipeline/Pipeline.groovy
GiriDandu: OK, I will remove it

src/main/groovy/deploydb/models/Pipeline/Pipeline.groovy
rtyler: I really do not believe this will work for configuring things. Whether we need an in-memory relation is a separate issue for us to deal with from this object. I don't see how Jackson can deserialize the desired YAML configuration file for this `Map` value

GiriDandu: For 0.1.0, can we change the configuration for pipeline, just to contain the ident of the enviroments and remove the promotions?

rtyler: I believe I understand where my confusion was, feel free to reference fuller namespaces in the comments to point out if I'm being dumb :) e..g `pipelines.Environment`

Provided the `pipelines.Environment` model allows for a null/non-existent list of promotions I think this should be fine

rtyler:
Looks like many of @maheshkelkar's comments were on commits instead of the pull request. Either way this isn't yet in a mergeable state

GiriDandu:
@maheshkelkar i will fix the PipelineSteps.groovy to make the text consistent

GiriDandu:
@rtyler why this not a mergeable state?  The latest changes takes care of all your comments from before

rtyler:
@GiriDandu "We can’t automatically merge this pull request. " which means there would be a conflict if we were to attempt to merge it. Please sync with the latest `upstream/master`

GiriDandu:
I will have one comment for all @maheshkelkar queries:

- pipeline directory is needed to isolate Environment and pipeline
- concurrent hash map is not required because it's read only data once the config is loaded. If we decide t o make config updatable on live machine, then you can change to concurrent hash map.
- Even if we initialize the promotions to [], it will not help during JSON deserialization and we will get compile error

Issue #48, opened 2015-02-17
Update Deployment Model to include Service, Promotion Status for the Artifact deployed


GiriDandu:
Deployment should serialize a "service" value as well as "environment"

maheshkelkar:
I am proposing following Mapper classes for Deployment

/* In memory object */
Deployment {
    Artifact artifact //reference
    Environment environment // reference
    Service service // reference
    State state
    PromotionStatus { ident, State, infoUrl)
}

/* Reading & writing to Db */
DeploymentDbMapper {
    Artifact artifact //reference OneToOne with Artifact
    String environment
    String service
}
why do we need this?
- Artifact reference is looked by hibernate; But, for conversions between ServiceId & Service, EnvironmentId & Environment object, we need this mapper.


/* JSON response */
DeploymentResponseMapper {
    Artifact artifact //reference
    String environment
    String service
    State state
    PromotionStatus { ident, State, infoUrl)
}
Why do we need this?
We dont want to expose entire Service or Environment object, so we need this to pick and choose what to expose.

GiriDandu:
Service, Environment and Promotion will be used in other objects in Webhooks, Flow etc. Can we make this generic. If its non trivial, then we can use separate mapper objects

maheshkelkar:
I deleted my comment about mapping. For the timebeing lets store serviceIdent & envinronmentIdent as String inside the Deployment object; we will perform in memory lookups for all the operations. Later we will sync up on how do this with more magic :)

PR #80, opened 2015-03-02
Add service, promotionResults SET, Status to Deployment

maheshkelkar:
Added PromotionResults model to contain Promotion Results - notably Status, infoUrl, etc. Added new Status enumeration. Used OneToMany and ManyToOne annotations to manage bi-directional references between Deployment (parent) and PromotionResults (children). Added Spock tests. 

References #48


features/api/deployment/reading.feature
```diff
@@ -24,23 +24,20 @@ Feature: Deployment READ APIs
...
-          "ident" : "jenkins-smoke",
-          "state" : "IN_PROGRESS",
-          "resultsUrl : ""
-        }]
+        "status" : "STARTED",
+        "promotionResultSet" : [{
+          "id" : 1,
+          "promotion" : "jenkins-smoke",
+          "status" : "IN_PROGRESS",
+          "infoUrl" : "",
```
rtyler: An empty value should be null, not an empty string IMO

features/api/deployment/reading.feature
```diff
@@ -24,23 +24,20 @@ Feature: Deployment READ APIs
...
-          "state" : "IN_PROGRESS",
-          "resultsUrl : ""
-         },
-         {
-          "ident" : "jenkins-smoke",
-          "state" : "IN_PROGRESS",
-          "resultsUrl : ""
-        }]
+        "status" : "STARTED",
+        "promotionResultSet" : [{
```
rtyler: Can this key just be changed to `promotions`?

features/api/deployment/updating.feature
```diff
@@ -14,10 +13,9 @@ Feature: Deployment UPDATE APIs
         "status" : "STARTED"
       }
     """
-    Then the response should be 200
+    Then the response should be 204
```
rtyler: I think this is fine for now, we might find it useflu to return the modifide Deployment object in the future which would change this to a 200 FWIW

src/main/groovy/deploydb/Status.groovy
```diff
@@ -0,0 +1,7 @@
+package deploydb
+
+/**
+ * Status enumeration
```
rtyler: This JavaDoc should explain more what the status is used for, I can already read that it's a status enumeration :P

src/main/groovy/deploydb/mappers/DeploymentMapper.groovy
```diff
@@ -1,16 +1,17 @@
...
+ * Simple Jackson mapper class to deserialize
+ * deployment update requests properly
  */
-class DeploymentMapper {
-    @JsonProperty
-    Long artifactId
-
-    @JsonProperty
-    String environment
+@Entity
```
rtyler: :+1:

src/main/groovy/deploydb/models/Deployment.groovy
```diff
@@ -45,6 +55,14 @@ class Deployment extends AbstractModel {
     Flow flow
 
     /**
+     * Add Promotion Result to collection
+     */
+    void addPromotionResult(PromotionResult promotionResult) {
```
rtyler: void doesn't seem like the right kind of response value for this. Will this have the potential to throw an exception? Is `add()` also a void call? If it's not perhaps we just return the return from `promotionResultSet.add()`

src/main/groovy/deploydb/models/PromotionResult.groovy
```diff
@@ -0,0 +1,92 @@
...
+import javax.persistence.EnumType
+import javax.persistence.FetchType
+import javax.persistence.JoinColumn
+import javax.persistence.ManyToOne
+import javax.persistence.Table
+import org.hibernate.validator.constraints.NotEmpty
+
+
+/**
+ * Representation class for the concept of a Deployment
```
rtyler: Copy paste failure >_<

src/main/groovy/deploydb/models/PromotionResult.groovy
```diff
@@ -0,0 +1,92 @@
...
+    @Column(name="promotion", nullable=false)
+    @JsonProperty(value = "promotion")
+    String promotionIdent
+
+    @Column(name="status", nullable=false)
+    @Enumerated(EnumType.ORDINAL)
+    @JsonProperty
+    Status status = Status.NOT_STARTED
+
+    @Column(name="infoUrl")
```
rtyler: This should be nullable, we're not always going to have an infoUrl.

src/main/groovy/deploydb/resources/DeploymentResource.groovy
```diff
@@ -1,66 +1,92 @@
...
+    @UnitOfWork
+    @Timed(name = "get-requests")
+    List<Deployment> getAll(
+            @QueryParam("pageNumber") @DefaultValue("0") IntParam pageNumber,
+            @QueryParam("perPageSize") @DefaultValue("20") IntParam perPageSize) {
+
+        /**
+         * Return empty list, if pageNumber and perPageSize is 0
+         */
+        if (pageNumber.get() == 0 && perPageSize.get() == 0) {
```
rtyler: Please file a ticket to refactor this. For all our paginated APIs we should add a pagination validator annotation or something like that so we don't have to repeat code everywhere

src/main/groovy/deploydb/resources/DeploymentResource.groovy
```diff
@@ -69,4 +95,28 @@ public class DeploymentResource {
...
+    void updateDeployment(@PathParam('id') LongParam deploymentId,
+                          @Valid DeploymentUpdateMapper deploymentUpdateMapper) {
+        Deployment deploy = this.dao.get(deploymentId.get())
+
+        if (deploy == null) {
+            throw new WebApplicationException(Response.Status.NOT_FOUND)
+        }
+
+        /**
+         *  FIXME: valid deploymentUpdateMapper.status are (Started, Completed, Failed)
```
rtyler: Please file issues and link to those issues in FIXME comments

src/main/groovy/deploydb/Status.groovy
rtyler: Also! If we're storing the position value from the enumeration (e.g. `IN_PROGRESS` gets stored as `1`) in the database this needs to be CLEARLY commented so that no values are ever removed and only added, otherwise we could fuck up our data at some point

src/main/groovy/deploydb/resources/DeploymentResource.groovy
maheshkelkar: We already have one for artifacts, we will just use it for that

features/api/deployment/reading.feature
maheshkelkar: Its null by default, the cucumber test is setting it to ""; if it sets it to null; it will show up as null

features/api/deployment/reading.feature
maheshkelkar: Awww, it took me real long while to get Set working; let me commit this and then I can create seperate ticket for followon effort

rtyler: `@JsonProperty(value='promotions')`

features/api/deployment/reading.feature
rtyler: Then please change the underlying code to create a null value so this makes sense from a doc perspective plz

features/api/deployment/reading.feature
maheshkelkar: Sorry, I misunderstood. Will change it and push again

maheshkelkar:
@rtyler do we have your go ahead to merge?

rtyler:
@maheshkelkar there's a couple notes which still haven't been addressed

src/main/groovy/deploydb/resources/DeploymentResource.groovy
maheshkelkar: I have added #83 to take care of these fixme(s)

maheshkelkar:
@rtyler sorry I see only 3 open comments; 1. you said its fine now, 2. thumbs up, 3. fixme issue is file. can you please point me to notes which are not addressed?

features/api/deployment/reading.feature
maheshkelkar: done

features/api/deployment/updating.feature
maheshkelkar: ok

src/main/groovy/deploydb/Status.groovy
maheshkelkar: done

src/main/groovy/deploydb/models/Deployment.groovy
maheshkelkar: done. changed signature.

src/main/groovy/deploydb/resources/DeploymentResource.groovy
maheshkelkar: #32 is updated to fix this for both artifacts & deployments

Issue #83, opened 2015-03-03
Incorporate workflow for all the internal and external triggers


maheshkelkar:
Implement or weave the pieces (artifact, deployment, flow, etc) of the workflows documented here:
http://hackers.lookout.com/deploydb/workflow.html

PR #86, opened 2015-03-05
Added WorkFlow object to piece all the objects together

maheshkelkar:
WorkFlow executes all the documented internal workflows

References #83

### DIFF
--- a/features/api/artifact/creating.feature
+++ b/features/api/artifact/creating.feature
@@ -9,6 +9,10 @@ Feature: Artifact CREATE APIs
 
     The creation of an artifact requires group, name, version and sourceUrl.
 
+    Given promotions are configured
+    And an environment is configured
+    And a pipeline is configured
+    And a service is configured
     When I POST to "/api/artifacts" with:
     """
       {
@@ -31,6 +35,7 @@ Feature: Artifact CREATE APIs
       }
     """
 
+
   @error
   Scenario: The request should fail when creating an Artifact without a group name
 
@@ -43,6 +48,7 @@ Feature: Artifact CREATE APIs
       }
     """
     Then the response should be 422
+
 
   @error
   Scenario: The request should fail when creating an Artifact without a name
@@ -57,6 +63,7 @@ Feature: Artifact CREATE APIs
     """
     Then the response should be 422
 
+
   @error
   Scenario: The request should fail when creating an Artifact without a version
 
@@ -69,6 +76,7 @@ Feature: Artifact CREATE APIs
       }
     """
     Then the response should be 422
+
 
   @error
   Scenario: The request should fail when creating an Artifact without a sourceUrl
@@ -83,6 +91,7 @@ Feature: Artifact CREATE APIs
     """
     Then the response should be 422
 
+
   @error
   Scenario: Creating an artifact with too large of a group name
 
@@ -90,6 +99,7 @@ Feature: Artifact CREATE APIs
     And the group is over 8192 characters
     When I POST to "/api/artifacts"
     Then the response should be 422
+
 
   @error
   Scenario: Creating an artifact with too large of a name
@@ -99,6 +109,7 @@ Feature: Artifact CREATE APIs
     When I POST to "/api/artifacts"
     Then the response should be 422
 
+
   @error
   Scenario: Creating an artifact with too large of a version
 
@@ -107,6 +118,7 @@ Feature: Artifact CREATE APIs
     When I POST to "/api/artifacts"
     Then the response should be 422
 
+
   @error
   Scenario: Creating an artifact with too large of a sourceUrl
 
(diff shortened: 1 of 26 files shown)


features/api/artifact/creating.feature
```diff
@@ -114,3 +126,76 @@ Feature: Artifact CREATE APIs
     And the sourceUrl is over 8192 characters
     When I POST to "/api/artifacts"
     Then the response should be 422
+
+
+  @error
```
GiriDandu: Do you want to cover the error cases for model (service, pipeline, environment) in their create feature files. This is only the artifact created.feature

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -0,0 +1,292 @@
...
+     */
+    void triggerArtifactCreated(models.Artifact artifact_) {
+
+        /* Persist the Artifact */
+        models.Artifact artifact = this.artifactDAO.persist(artifact_)
+
+        /* Lookup Service */
+        models.Service service = this.serviceRegistry.getAll().find() { models.Service service ->
+            service.artifacts.find() { String artifactIdent ->
+                artifactIdent == (artifact.group + ":" + artifact.name)
```
GiriDandu: Is this == or =?

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -0,0 +1,292 @@
...
+            }
+        }
+        if (service == null) {
+            /* Artifact is left hanging */
+            String serviceIdent = artifact.group + ":" + artifact.name
```
GiriDandu: Why the serviceIdent being created here in exception handling

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -0,0 +1,292 @@
...
+        List<models.pipeline.Pipeline> pipelines = service.getPipelines().collect() { String pipelineIdent ->
+            this.pipelineRegistry.get(pipelineIdent)
+        }
+        if (pipelines.isEmpty() || pipelines.contains(null)) {
+            throw new WebApplicationException(Response.Status.NOT_FOUND)
+        }
+
+        /* Get all environments */
+        List<deploydb.models.Environment> environments = []
+        List<models.Promotion> pipePromotions = []
```
GiriDandu: Nit pick, can full expand the names to "pipelinePromotions" and "pipelineEnvironments"

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -0,0 +1,292 @@
...
+            if (flow.status != Status.FAILED) {
+                flow.status = Status.SUCCESS
+            }
+            return
+        }
+
+        /* Mark Deployment as ready for deploying */
+        deploy.status = Status.CREATED
+
+        println "***MVK: triggerDeploymentCreated: deploy: ${deploy}"
```
GiriDandu: Remove the println

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -0,0 +1,292 @@
...
+     */
+    void triggerDeploymentCreated(models.Flow flow) {
+
+        /* Ignore if flow is marked for failed */
+        if (flow.status == Status.FAILED) {
+            return
+        }
+
+        /* Make Sure no deployment in progress */
+        models.Deployment deploy = flow.getDeployments().find() { d ->
```
GiriDandu: full expand the name to "deployment", no "deploy"

features/api/artifact/creating.feature
maheshkelkar: These are errors encountered on artifact creates; hence I think we should leave it there

src/main/groovy/deploydb/WorkFlow.groovy
maheshkelkar: ==, when the condition succeeds, then find() returns

src/main/groovy/deploydb/WorkFlow.groovy
maheshkelkar: Debugging code. I will remove it

features/api/artifact/creating.feature
```diff
@@ -9,6 +9,10 @@ Feature: Artifact CREATE APIs
 
     The creation of an artifact requires group, name, version and sourceUrl.
 
+    Given promotions are configured
```
rtyler: This is pretty verbose and unnecessary for the documentation of this API. I suggest writing `Given DeployDB is configured` and then calling the same underlying methods that these four steps are calling instead

features/api/artifact/creating.feature
rtyler: :+1:  I wanted to tag expected error cases with `@error` so it would be easier for us to catalogue them in the future.

features/api/artifact/creating.feature
```diff
@@ -114,3 +126,76 @@ Feature: Artifact CREATE APIs
     And the sourceUrl is over 8192 characters
     When I POST to "/api/artifacts"
     Then the response should be 422
+
+
+  @error
+  Scenario: Creating a new artifact with missing promotion fails
```
rtyler: I don't understand this scenario. Why is this a failure? What does the API user care about promotions whatsoever?

features/api/artifact/creating.feature
```diff
@@ -114,3 +126,76 @@ Feature: Artifact CREATE APIs
...
+        "name" : "cukes",
+        "version" : "1.2.345",
+        "sourceUrl" : "http://example.com/cucumber.jar"
+      }
+    """
+    Then the response should be 404
+
+
+  @error
+  Scenario: Creating a new artifact with missing environment fails
```
rtyler: Similar questions for these other error cases

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -0,0 +1,292 @@
...
+     *
+     * Creates deployments and flows for the given artifact, based on the
+     * information provided in the configured models
+     *
+     * @param artifact
+     */
+    void triggerArtifactCreated(models.Artifact artifact_) {
+
+        /* Persist the Artifact */
+        models.Artifact artifact = this.artifactDAO.persist(artifact_)
```
GiriDandu: Move this to the end, once all the configuration are correct as a precondition before artifact is created

features/api/artifact/creating.feature
maheshkelkar: Giri and I discussed both the approaches and we chose this approach for 2 reasons:
1. it explains what models are to be configured
2. it helps with testing error scenarios (see comment below)

features/api/artifact/creating.feature
maheshkelkar: If one of the promotion defined by service or pipeline is not available, then artifact create should fail. This helped me test all the paths of "triggerArtifactCreated"

features/api/artifact/creating.feature
maheshkelkar: I agree that we will validate the config at the load time. But, this sort of tests ensure that if models get accepted due to misconfiguration, we will still catch them.

we would not create artifacts if one of the models (i.e. environment in this case)  is not available

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -0,0 +1,291 @@
...
+     *
+     * This is an internal trigger fired when deployment is verified (success/failed)
+     * OR when artifact is created
+     *
+     * @param flow
+     */
+    void triggerDeploymentCreated(models.Flow flow) {
+
+        /* Ignore if flow is marked for failed */
+        if (flow.status == Status.FAILED) {
```
GiriDandu: A log would help here

maheshkelkar: I think, we need activity log strategy. A seperate issue to add it everywhere

Issue #45, opened 2015-02-13
How DeployDb loads yaml-based/configuration-driven models


maheshkelkar:
The Service, Environment, Promotions, and Pipeline are all configuration-driven models. 

During launch, the DeployDb needs to read the YAML file for each of these models and create their instances.

maheshkelkar:
@rtyler for location of all yaml files, do we assume a fixed path (e.g. /etc/deploydb/resource) OR read it from the DeployDb specific configuration file (i.e. deploydb.example.yml)

rtyler:
@maheshkelkar the location of service files should be read from the "primary" configuration file in my opinion, @polynomial can correct me if he thinks otherwise

maheshkelkar:
@rtyler @polynomial @GiriDandu 
To load configuration files during launch, I propose the following:

- New attribute in the primary configuration file
e.g. deploydb.example.yml OR deploydb.cucumber.yml

- The new attribute is as follows:
 "configFilesPath: "/etc/deploydb"

- During launch DeployDbApp will read Service, Environment, Promotion, Pipeline and webhooks from their own subdirectories to populate models:
E.g. 
- services/faas.yml (i.e. /etc/deploydb/services/faas.yml) - for services
- environments/integ.yml (i.e. /etc/deploydb/environment/integ.yml) - for environments
etc

GiriDandu:
@maheshkelkar I like this idea in general. Instead of "configFilesPath" can we change this "configDirectory"

polynomial:
Agree with @rtyler about the initial location read from the primary config. I'm hoping the rest can be:
* read with just a signal, not a restart
* on failure to validate, should not break the service
* do not need to be explicitly defined anywhere (it should recurse the configFilesPath directory and load everything ending in .yml)

rtyler:
@maheshkelkar I also like `configDirrectory` as the key name in the primary configuration. Dropwizard already provides a command line flag for specifying where that primary configuration is so that should be enough.

To @polynomial's requests, items one and two we shouldn't address right now, but the configuration file handling code should be written such that we can call it multiple times thorughout the lifetime of the application (boot, and then multiple times later).

maheshkelkar:
@rtyler, @GiriDandu configDirectory it is. I have experimented in defining this as an attribute in deploydb.yml; I assume dropwizard should automagically allow that into command line option. 

[...]

PR #93, opened 2015-03-10
Added configDirectory parameter to load yaml config during launch

maheshkelkar:
Added a new parameter, iterating through directories and loading configuration,
configuration validation and sanitization, etc

References #45


deploydb.cucumber.yml
```diff
@@ -2,6 +2,12 @@ database:
   driverClass: org.h2.Driver
   user: nobody
   url: "jdbc:h2:mem:cucumber"
+
+# DeployDB base configuration directory (cannonical path)
+# YAML configuration files for models - services, environments, pipelines,
+# webhooks and promotions - should be stored in respective sub-directories
+#configDirectory: config
+
```
GiriDandu: Why do you have this commented out. 
You want to remove it from this pull request?

src/main/groovy/deploydb/DeployDBApp.groovy
```diff
@@ -16,11 +16,12 @@ import org.joda.time.DateTimeZone
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 
+
 class DeployDBApp extends Application<DeployDBConfiguration> {
     private final ImmutableList models = ImmutableList.of(
             models.Artifact, models.Deployment,
             models.PromotionResult, models.Flow)
-    private final Logger logger = LoggerFactory.getLogger(DeployDBApp.class)
+    private static final Logger logger = LoggerFactory.getLogger(DeployDBApp.class)
```
GiriDandu: Why was this changed from final to static.  We have DeployDBApp available everywhere, why do you need to make this class scope?

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -47,102 +55,295 @@ public class WorkFlow {
...
+        models.Webhook.Webhook tmpWebhook
+
+        /* Load promotions */
+        String promotionsDirName = baseConfigDirName + "/promotions"
+        File promotionsDirectory = new File(promotionsDirName);
+        if (promotionsDirectory.exists() && promotionsDirectory.isDirectory()) {
+            logger.info("Loading promotions from directory: ${promotionsDirectory.getCanonicalPath()}");
+
+            /* Skip everything but yaml file */
+            promotionsDirectory.eachFileMatch(FileType.FILES, ~/^.*?\.yml/) { File modelFile ->
```
GiriDandu: Can you not parameterize this functionality, this code appears few times below

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -47,102 +55,295 @@ public class WorkFlow {
...
+                try {
+                    models.Promotion promotion = this.promotionLoader.load(modelFile)
+                    promotion.ident = this.promotionLoader.getIdent(modelFile.name)
+                    tmpPromotionRegistry.put(promotion.ident, promotion)
+                    logger.debug("Loaded promotions model: ${promotion.ident}")
+
+                } catch (all) {
+                    logger.info("Failed to load promotion model from ${modelFile.name}")
```
GiriDandu: In exception handling code, the model which are already configured will be initialized in half baked state. If this function is executed more than once, then there will be multiple entries and it is a problem. This is especially a big problem for webhooks, where we may send notifies more than once for a single configuration

src/test/groovy/deploydb/WorkFlowSpec.groovy
```diff
@@ -13,3 +13,172 @@ class WorkFlowSpec extends Specification {
...
+        given:
+        createPromotionConfigFile()
+        createEnvironmentConfigFile()
+        createPipelineConfigFile()
+        createServiceConfigFile()
+        createWebhookConfigFile()
+        workFlow.loadConfigModels(baseCfgDirName)
+
+        expect:
+        workFlow.promotionRegistry.getAll().size() == 1
```
GiriDandu: Why is only ident being checked here. Why not check the content of each registry.

deploydb.cucumber.yml
maheshkelkar: We dont need this config for cucumber test. But, I wondered if we need to leave this in there as an FYI

src/main/groovy/deploydb/DeployDBApp.groovy
maheshkelkar: I tried to make it consistent with usage in dropwizard app. Also, I dont think it is class scope, static will only guarantee only 1 instance for class

src/main/groovy/deploydb/WorkFlow.groovy
maheshkelkar: We collect all of these models into temporary registry. Once we validate all the registries in the end and there are NO active flows, then we will overwrite the existing registries. So we are ensuring that there is no conflict

src/test/groovy/deploydb/WorkFlowSpec.groovy
maheshkelkar: We have test code that validates this in their respective spock tests. Here, we are only ensuring that expected number and ident model is populated in the application/workflow

src/main/groovy/deploydb/WorkFlow.groovy
GiriDandu: Ah, I see in the code at the end of the function, you copy everything from temp registry to WorkFlow object, that should prevent duplicate entries. Although we have allocated objects in the process before exception handling, we expect the objects cleaned up by GC

src/test/groovy/deploydb/WorkFlowSpec.groovy
GiriDandu: Then I don't know why you need this test because we have the same identity tests in model spock test also. What's more this function is going to do?

GiriDandu:
:+1:

Issue #95, opened 2015-03-10
Add REST API to GET Results of a promotion associated with the deployment


maheshkelkar:
Also, update deployment REST APIs to return updated deployment object

PR #96, opened 2015-03-10
Define REST API to GET results of a promotion assoc with the deployment

maheshkelkar:
References #95

### DIFF
--- a/features/api/deployment/updating.feature
+++ b/features/api/deployment/updating.feature
@@ -5,7 +5,7 @@ Feature: Deployment UPDATE APIs
 
 
   @freezetime @wip
-  Scenario: Updating a deployment with a status STARTED
+  Scenario: Updating a deployment with a status COMPLETED
 
     Given there is a deployment
     When I PATCH "/api/deployments/1" with:
@@ -43,7 +43,7 @@ Feature: Deployment UPDATE APIs
 
 
   @freezetime @wip
-  Scenario: Updating a deployment with a status COMPLETED
+  Scenario: Updating a deployment with a status FAILED
 
     Given there is a deployment
     When I PATCH "/api/deployments/1" with:
--- a/features/api/deployment/updating.feature
+++ b/features/api/deployment/updating.feature
@@ -4,6 +4,7 @@ Feature: Deployment UPDATE APIs
   I should be able to update existing deployments in the system
 
 
+  @freezetime @wip
   Scenario: Updating a deployment with a status STARTED
 
     Given there is a deployment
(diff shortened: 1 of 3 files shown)


features/api/deployment/updating.feature
```diff
@@ -4,6 +4,7 @@ Feature: Deployment UPDATE APIs
   I should be able to update existing deployments in the system
 
 
+  @freezetime @wip
   Scenario: Updating a deployment with a status STARTED
```
GiriDandu: Can you please be more expressive. What is the reason for this "update"

GiriDandu: Can you please be more expressive. State the reason for updating, why did we get this trigger

features/api/deployment/updating.feature
```diff
@@ -13,9 +14,35 @@ Feature: Deployment UPDATE APIs
         "status" : "COMPLETED"
```
GiriDandu: Are we updating status with STARTED or COMPLETED?

features/api/promotion/reading.feature
```diff
@@ -46,3 +46,36 @@ Feature: Promotion READ APIs
...
+    Given there is a deployment
+    When I GET "/api/deployments/1/promotions/1"
+    Then the response should be 200
+    And the body should be JSON:
+    """
+      {
+        "id" : 1,
+        "promotion" : "jenkins-smoke",
+        "status" : "STARTED",
+        "infoUrl" : null,
```
GiriDandu: Why the infoUrl "null" here.  For a success case, should we provide the correct infoUrl?

features/api/promotion/reading.feature
```diff
@@ -46,3 +46,36 @@ Feature: Promotion READ APIs
...
+        "createdAt" : "{{created_timestamp}}"
+      }
+    """
+
+
+  @wip
+  Scenario: Fetching Promotion result for a Deployment that doesn't exist
+
+    When I GET "/api/deployments/1/promotions/1"
```
GiriDandu: This is deployment that doesn't exist. More interesting case would be deployment is valid but promotion doesn't exist

features/api/deployment/updating.feature
maheshkelkar: This part is already reviewed and approved and not changed in this commit. 

IMO scenario(s) are very specific and succinct. What if I update the feature "description" (at the top) as follows:
The deployDB maintains the status or information about the deployments in central repository. These REST APIs allow the clients to update the status or information of existing deployments and provide up-to-date view for the benefit of administrator.

features/api/deployment/updating.feature
maheshkelkar: fixed the text in scenario

features/api/promotion/reading.feature
maheshkelkar: Currently promotion result is in STARTED state, so I dont have infoUrl. In general infoUrl is optional and can be null

features/api/promotion/reading.feature
maheshkelkar: The next scenario already covers this:

Scenario: Fetching a result for a Promotion which is not associated with the Deployment

features/api/deployment/updating.feature
GiriDandu: Just to make sure I understand this better.  In what circumstance the administrator would change the state of the deployment.  To retry the failed deployment again. It's not clear (at least to me) why you need this during "normal/working"  system

features/api/promotion/reading.feature
GiriDandu: ok, i get it

features/api/deployment/updating.feature
maheshkelkar: Administrator is simply looking at the console and tracking all the deployments. 

The client i.e. orchestrator (rundeck) is updating the status of deployment to report the progress of deployment.

GiriDandu: OK, I understand now

GiriDandu:
:+1:

Issue #77, opened 2015-02-25
201 CREATED responses should include LOCATION header of newly created resource which can be GET'd later.


tommysullivan:
If not solved generally, then please solve for POST to deployments/:deploymentId/promotions so that I can link back to the promotion resource once pa-portal creates it. This also enables our blackbox cukes to validate that promotion occurred and a promotion artifact was produced and that its status is as expected.

maheshkelkar:
@tommysullivan Let me know if this satisfies the requirement?

Create Artifact:

```
thq-m-mkelk01:deploydb-0217 mkelkar$ curl -v -X POST -H  "Content-Type: application/json" http://localhost:8080/api/artifacts -d '{"group" : "basic.group.1", "name": "bg1", "version" : "1.2.3.4", "sourceUrl" : "http://example.com/cucumber.jar"}'
* Hostname was NOT found in DNS cache
*   Trying ::1...
* Connected to localhost (::1) port 8080 (#0)
> POST /api/artifacts HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8080
> Accept: */*
> Content-Type: application/json
> Content-Length: 114
>
* upload completely sent off: 114 out of 114 bytes
< HTTP/1.1 201 Created
< Date: Tue, 10 Mar 2015 19:41:37 GMT
< Location: http://localhost:8080/api/artifacts/1  <<<<<<<<<<<<<<<<<<<<
< Content-Type: application/json
< Vary: Accept-Encoding
< Content-Length: 150
<
* Connection #0 to host localhost left intact
{"id":1,"createdAt":"2015-03-10T19:41:37.123Z","group":"basic.group.1","name":"bg1","version":"1.2.3.4","sourceUrl":"http://example.com/cucumber.jar"}
```

Promotion Results add to deployment:

```
thq-m-mkelk01:deploydb-0217 mkelkar$ curl -v -X POST -H  "Content-Type: application/json" http://localhost:8080/api/deployments/1/promotions -d '{"name": "basicPromo", "status" : "SUCCESS", "infoUrl" : "http://local.lookout.com/jenkins/job-id/2/results"}'
* Hostname was NOT found in DNS cache
*   Trying ::1...
* Connected to localhost (::1) port 8080 (#0)
> POST /api/deployments/1/promotions HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8080
> Accept: */*
> Content-Type: application/json
> Content-Length: 109
>
[...]

PR #98, opened 2015-03-11
Return location header in created response and implemented API for promotion results

maheshkelkar:
Location header will be added to artifact and promotion result created response. Other
REST APIs updated as defined in issue 95

References #77

### DIFF
--- a/src/main/groovy/deploydb/resources/ArtifactResource.groovy
+++ b/src/main/groovy/deploydb/resources/ArtifactResource.groovy
@@ -6,6 +6,8 @@ import deploydb.WorkFlow
 import io.dropwizard.jersey.params.IntParam
 import io.dropwizard.jersey.params.LongParam
 import io.dropwizard.hibernate.UnitOfWork
+
+import javax.servlet.http.HttpServletRequest
 import javax.validation.Valid
 import javax.ws.rs.Consumes
 import javax.ws.rs.DefaultValue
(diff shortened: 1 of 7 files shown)


src/main/groovy/deploydb/resources/ArtifactResource.groovy
```diff
@@ -38,11 +40,13 @@ public class ArtifactResource {
...
     @UnitOfWork
     @Timed(name='put-requests')
-    Response createArtifact(@Valid Artifact artifact) {
+    Response createArtifact(@Context HttpServletRequest request,
+                            @Valid Artifact artifact) {
 
         this.workFlow.triggerArtifactCreated(artifact)
 
-        return Response.status(201).entity(artifact).build()
+        String createdUri = request.getRequestURL() + "/${artifact.id}"
```
GiriDandu: Can you avoid local variable createdUri. "request.getRequestURL()+"/${artifact.id}".toURI()

src/main/groovy/deploydb/resources/DeploymentResource.groovy
```diff
@@ -177,5 +183,38 @@ public class DeploymentResource {
         } else if (promotionResultAddMapper.status == Status.FAILED) {
             this.workFlow.triggerPromotionFailed(deploy)
         }
+
+        /**
+         * Prep the response
+         */
+        String createdUri = request.getRequestURL() + "/${promotionResult.id}"
```
GiriDandu: Same comment as before, please remove temporary variable

src/main/groovy/deploydb/resources/ArtifactResource.groovy
maheshkelkar: I think extra variable makes it look saner, otherwise we would need:
return Response.created((request.getRequestURL() + "/${artifact.id}").toURI).entity(artifact).build()

GiriDandu:
:+1:

Issue #99, opened 2015-03-11
Add restart logic and DB persistence


maheshkelkar:

maheshkelkar:
@rtyler I think I am going to use this issue to make deployDB work when persistence is configured. And test the "db" commands. I am assuming we will create a separate to address the challenges incurred when running with multiple instances of deployDB and database.

rtyler:
This strikes me as though it should be two separate tickets.

PR #100, opened 2015-03-12
Removed flyway instatiation in app, it will be driven by db command

maheshkelkar:
For deploydb.example.yml, use persistent h2 instead of in-memory

References #99

### DIFF
--- a/deploydb.example.yml
+++ b/deploydb.example.yml
@@ -1,7 +1,7 @@
 database:
   driverClass: org.h2.Driver
   user: nobody
-  url: "jdbc:h2:~/deployDB.example.file"
+  url: "jdbc:h2:./deployDB.example.db"
 
 # DeployDB base configuration directory (cannonical path)
 # YAML configuration files for models - services, environments, pipelines,
--- a/deploydb.example.yml
+++ b/deploydb.example.yml
@@ -1,7 +1,7 @@
 database:
   driverClass: org.h2.Driver
   user: nobody
-  url: "jdbc:h2:mem"
+  url: "jdbc:h2:~/deployDB.example.file"
 
 # DeployDB base configuration directory (cannonical path)
 # YAML configuration files for models - services, environments, pipelines,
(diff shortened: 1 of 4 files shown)


rtyler:
Will you update the README to reflect the "proper" setup steps?

src/main/groovy/deploydb/models/Artifact.groovy
```diff
@@ -82,4 +82,11 @@ class Artifact extends AbstractModel {
         return Objects.hash(this.id, this.group, this.name,
                             this.version, this.sourceUrl)
     }
+
+    @Override
```
rtyler: This change has nothing to do with this commit and pull request

src/main/groovy/deploydb/models/Deployment.groovy
```diff
@@ -131,7 +131,7 @@ class Deployment extends AbstractModel {
     @Override
     String toString() {
         return "id: ${id}, environment: ${environmentIdent}, service: ${serviceIdent}, " +
-                "status: ${status}, " + "promotionResultSet: ${promotionResultSet}, " +
+                "status: ${status}, promotionResultSet: ${promotionResultSet}, " +
```
rtyler: This change also has nothing to do with this change

src/main/groovy/deploydb/DeployDBApp.groovy
```diff
@@ -89,14 +87,6 @@ class DeployDBApp extends Application<DeployDBConfiguration> {
     @Override
     public void run(DeployDBConfiguration configuration,
                     Environment environment) {
-        /* Execute migrations */
-        Flyway flyway = new Flyway()
```
rtyler: :+1:

deploydb.example.yml
```diff
@@ -1,7 +1,7 @@
 database:
   driverClass: org.h2.Driver
   user: nobody
-  url: "jdbc:h2:mem"
+  url: "jdbc:h2:~/deployDB.example.file"
```
rtyler: Can you make this default to `./deploydb.example.db` instead, so when people run the example code they're not having files dumped in their home directory unexpectedly?

maheshkelkar: ok

src/main/groovy/deploydb/models/Artifact.groovy
maheshkelkar: But, right thing to do, hence rolled it in :)

src/main/groovy/deploydb/models/Deployment.groovy
maheshkelkar: But, right thing to do, hence rolled it in :)

PR #104, opened 2015-03-13
Added instructions for how to run the deployDb application

maheshkelkar:
Only contains document updates


rtyler:
These instructions are much more oriented towards a developer who is already working with DeployDB, which is fine but I'll write up some additional instructions for taking the distribution package and running DeployDB from that

PR #109, opened 2015-03-18
Issue 107 deploydb whoas configuration support

GiriDandu:
Add whoasfactory and use it to build queue and runner
Use latest whoas library - 0.1.4
Remove unnecessary webhookmanager creation in webhooksteps

### DIFF
--- a/src/cucumber/groovy/step_definitions/WebhookSteps.groovy
+++ b/src/cucumber/groovy/step_definitions/WebhookSteps.groovy
@@ -21,6 +21,9 @@ import org.joda.time.DateTime
 import deploydb.WebhookManager
 import webhookTestServer.models.RequestWebhookObject
 
+def getClassVariable(klass, variable) {
+    return klass."$variable"
+}
 
 Given(~/^a webhook "(.*?)" configuration:$/) { String eventType, String configBody ->
 
@@ -110,14 +113,13 @@ And(~/there is a deployment in "(.*?)" state$/) { String deploymentState ->
          * Create sample promotionResult(s)
          */
         PromotionResult p1 = new PromotionResult("jenkins-smoke", Status.STARTED, null)
-        WebhookManager webhookManager = new WebhookManager()
         /**
          * Create deployment
          */
         Deployment d1 = new Deployment(a1,
                 "pre-prod",
                 "faas",
-                webhookManager.getMemberOfObject(Status, deploymentState))
+                getClassVariable(Status, deploymentState))
         d1.addPromotionResult(p1)
 
         /* Create a flow */
--- a/src/cucumber/groovy/step_definitions/WebhookSteps.groovy
+++ b/src/cucumber/groovy/step_definitions/WebhookSteps.groovy
@@ -21,10 +21,6 @@ import org.joda.time.DateTime
 import deploydb.WebhookManager
 import webhookTestServer.models.RequestWebhookObject
 
-def getClassVariable(klass, variable) {
-    return klass."$variable"
-}
-
 Given(~/^a webhook "(.*?)" configuration:$/) { String eventType, String configBody ->
 
     List<String> paths = getUrlPathFromWebhookConfigBody(configBody, eventType)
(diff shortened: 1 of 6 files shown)


src/main/groovy/deploydb/DeployDBConfiguration.groovy
```diff
@@ -3,6 +3,7 @@ package deploydb
 import io.dropwizard.Configuration
 import io.dropwizard.db.DataSourceFactory
 import io.dropwizard.flyway.FlywayFactory
```
maheshkelkar: It seems that FlywayFactory is defined in dropwizard and not the other way round

But, for whoas we are importing the dropwizard. As a result are we mandating use of Whoas with Dropwizard (or this configuration feature is only available when used with dropwizard)

src/main/groovy/deploydb/DeployDBConfiguration.groovy
```diff
@@ -49,5 +53,15 @@ class DeployDBConfiguration extends Configuration {
     void setConfigDirectory(String configDirectory) {
         this.configDirectory = configDirectory
     }
+
+    @JsonProperty("whoas")
+    WhoasFactory getWhoasFactory() {
```
maheshkelkar: I think we do not need any getter & setter routines (since this is in groovy).

src/main/groovy/deploydb/DeployDBConfiguration.groovy
GiriDandu: Whoas can be used without factory and configuration. There is no need to use factory to instantiate  whoas queue and runner. 
In case of deploydb (which uses dropwizard), its a convenient way to build queue and runner based on configuration.  
Your observation is correct, if we want to use factory, then we are requiring the clients of whoas to use dropwizard configuration.

src/main/groovy/deploydb/DeployDBConfiguration.groovy
GiriDandu: Sure i will remove it

src/cucumber/groovy/step_definitions/WebhookSteps.groovy
```diff
@@ -110,14 +113,13 @@ And(~/there is a deployment in "(.*?)" state$/) { String deploymentState ->
...
         PromotionResult p1 = new PromotionResult("jenkins-smoke", Status.STARTED, null)
-        WebhookManager webhookManager = new WebhookManager()
         /**
          * Create deployment
          */
         Deployment d1 = new Deployment(a1,
                 "pre-prod",
                 "faas",
-                webhookManager.getMemberOfObject(Status, deploymentState))
+                getClassVariable(Status, deploymentState))
```
rtyler: Since this is only used in one place, why not just dynamic dispatch directly? e.g. `Status."${deploymentState}"`

src/main/groovy/deploydb/WebhookManager.groovy
```diff
@@ -30,8 +28,21 @@ class WebhookManager implements Managed {
...
+     * Will be used only during unit tests
+     */
+    WebhookManager() { }
+
+    /**
+     *  Constructor to create based on configuration of queues and runner
+     *
+     * @param deployDBConfiguration
+     */
+    WebhookManager(DeployDBConfiguration deployDBConfiguration ) {
```
rtyler: Such verbose naming, why not just call the variable `config` :P

src/main/groovy/deploydb/WebhookManager.groovy
```diff
@@ -30,8 +28,21 @@ class WebhookManager implements Managed {
...
+
+        queue = deployDBConfiguration.getWhoasFactory().buildQueue()
```
rtyler: `queue = config.whoasFactory.buildQueue()`

src/main/groovy/deploydb/WebhookManager.groovy
GiriDandu: I strongly believe in self commenting code. So long names is what we should be using.

rtyler: Eh, I don't feel passionately enough about this to argue about this. We have a type system available and I find overly verbose variable names reduce readability, not improve it.

Issue #103, opened 2015-03-13
When used with MySql on AWS, Some SQL commands failed to execute 


maheshkelkar:
This problem was observed while testing DeployDb on EC2 instance and MySQL database on RDS instance in AWS.

Incompatibility between SQL syntax of H2 and MySql caused this problem.

Notablly, it failed to execute following commands in V3 of the migration:

SQL commands that worked on H2:
```
ALTER TABLE artifacts ALTER COLUMN groupName VARCHAR(8192);
ALTER TABLE artifacts ALTER COLUMN name VARCHAR(8192);
 
 /* No need for this index anymore */
DROP INDEX version_index;
 
CREATE UNIQUE INDEX artifact_version_index ON artifacts(groupName, name, version);
```

SQL commands that worked on MySql:
```
ALTER TABLE artifacts MODIFY COLUMN groupName VARCHAR(8192) NOT NULL;
ALTER TABLE artifacts MODIFY COLUMN name VARCHAR(8192) NOT NULL;
 
 /* No need for this index anymore */
DROP INDEX version_index ON artifacts;
 
CREATE UNIQUE INDEX artifactveridx ON artifacts ( groupName(50), name(50), version(50) );
```

The errors were as follows:
For ALTER TABLE:
```
Error executing statement at line 6: ALTER TABLE artifacts ALTER COLUMN groupName VARCHAR(8192)
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'VARCHAR(8192)' at line 1
```

For DROP INDEX:
```
Error executing statement at line 10: DROP INDEX version_index
```

For CREATE UNIQUE INDEX:
```
Specified key was too long; max key length is 767 bytes
Error executing statement at line 12: CREATE UNIQUE INDEX artifact_version_index ON artifacts(groupName, name, version)
```

maheshkelkar:
Giri & I investigated further in this sql-incompatibility between H2 and MySQL; we have 3 choices:
define separate sql db migration files (duplicate scripts, maintainability issues, etc)
Write a script or contribute to existing github project mysql2h2-converter (time consuming, requires running this script on sql before every use)
Use flyway Java migrations. Here I am making an assumption that Driver name (H2 or mysql) will be available and will help us use the DB appropriate command and generate the appropriate SQL command.

rtyler:
[...]

PR #114, opened 2015-03-21
Using flyway java(groovy) migrations for sql compatibility with diff DBs

maheshkelkar:
E.g. MySQL and H2 sql commands are different. Using flyway's java (i.e. groovy in our case)
migrations, instead of sql migrations, we can have tailormade sql commands for each db.

Also fixed the stubAppRunner (i.e. cucumber tests) to read and apply the database/flyway
config from yaml file (instead of hardcoding it)

References #103


build.gradle
```diff
@@ -51,6 +51,7 @@ dependencies {
 
     compile withSources('com.h2database:h2:1.3.+')
     compile withSources('joda-time:joda-time:2.6+')
+    compile 'mysql:mysql-connector-java:5.1.+'
```
rtyler: Why is this dependency being added? I don't believe this should be here.

src/main/groovy/db/migration/V1__create_artifacts_table.groovy
```diff
@@ -0,0 +1,42 @@
...
+
+import org.flywaydb.core.api.migration.jdbc.JdbcMigration
+import java.sql.Connection
+import java.sql.PreparedStatement
+
+
+/**
+ * Example of a Java-based migration.
+ */
+public class V1__create_artifacts_table implements JdbcMigration {
```
rtyler: `public` isn't necessary in Groovy

rtyler:
This is a good cut; I like reading code on the weekends :D

There is a lot of duplicated code between the statement preparation and the "h2" connector checking. Can you put a Groovy class in the `db.migration` package which would be `DeployDBMigration` or something like that which implements `JdbcMigration` and allows us to consolidate some of these common behaviors between our migrations?

We also need to implement `getChecksum` for [checksumkming and validation](
http://flywaydb.org/documentation/migration/java.html#Checksums_and_Validation) to make sure we can use Flyway's change detection. I think that `getChecksum` can just be implemented in every class to return the migration number (e.g. "3", "4", etc)

I would suggest making the API something like the following to make our migrations easier, I've not tested this code so YMMV :):

```groovy
@TypeChecked
abstract class DeployDBMigration implements JdbcMigration, MigrationChecksumProvider {
    protected Logger logger = LoggerFactory.getLogger(DeployDBMigration.class)

    /** Implementing classes responsible for providing a list of String commands */
    abstract List<String> prepareCommands(DatabaseMetadata metadata)

    @Override
    void migrate(Connection connection) {
        executeCommands(connection, prepareCommands(connection.metaData))
    }

    /** Tell me if I'm executing this migration on this driver */
    Boolean isH2(String driverName) {
        return driverName.matches("(?i)H2")
    }

    /** Execute the commands after preparing all the statements */
    private void executeCommands(Connection connection, List<String> commands) {
        commands.each { String -> command
            logger.info("Preparing statement for: ${command}")
            PreparedStatement statement = connection.prepareStatement(command)

            try {
                statement.execute()
            finally {
                statement.close()
            }
        }
    }
}
```


```groovy
class TylersFirstMigration extends DeployDBMigration {
    List<String> prepareCommands(DatabaseMetadata metadata) {
        if (isH2(metadata.driverName)) {
            return ["Some H2 Query"]
        }
        return []
    }
}
```

build.gradle
maheshkelkar: I think this includes the MySQL client java classes. Without this, DeployDB fails to find the com.MySql.Driver class.

src/main/groovy/db/migration/V1__create_artifacts_table.groovy
maheshkelkar: Yes, will take it out. Started with java code and later on converted to groovy code. Wasnt sure if flyway would like that.

build.gradle
rtyler: I get this, but this also requires that our specific version of the MySQL Connector is loaded into the classpath and bundled with *every* DeployDB artifact.

What I had imagined previously was that if somebody wanted to use another JDBC connector that they would need to ensure that the appropriate connector classes were installed on their machine(s) outside of the DeployDB artifact.  Since I wouldn't want to include every JDBC connecor client in the `.jar`

maheshkelkar: I tried that on AWS. I downloaded mySQ: connector jar file, setup classpath correctly. But, deploydb kept complaining that it could not locate or load the MySql.Driver class.

maheshkelkar: I can set this up again. But, I may need some pointers or help in what I might be doing wrong.

IMO this is no different than running client jar/library in your application. 

Since we are doing the same thing with H2, I thought it would be no different.

rtyler: The difference between the two is that H2 is the bare minimum default needed to actually run. Including it in the base distribution allows us to support pretty good/easy standalone installations of DeployDB for demonstration purposes.

I'll go ahead and merge this and then write up some documentation on how to install a JDBC connector library in such a way that it can be used from DeployDB without needing to include it

PR #119, opened 2015-03-25
Issue 106 webhook content type for events

GiriDandu:


rtyler:
There's not any actual documentation of what the value of the `Content-Type` header @orrc can expect in the cukes.

I suggest adding another step fpr checking headers on the payload, e.g.:

```cucumber
Then the request should have the headers:
  |  Name        |   Value                                           |
  | Content-Type |  application/vnd.deploydb.somethingsomething+json |
```

Constructing the step with Cucumber data tables I think will also make it much more reusable for documenting other headers too.

GiriDandu:
@rtyler ok that makes sense, I will add additional step to to verify content-type using data table

rtyler:
The string concatenation looks brittle but I don't have better suggestions right now

Issue #120, opened 2015-03-26
Define new REST API to reload configuration with an API call


maheshkelkar:
Related to #44

PR #121, opened 2015-03-26
Defined REST API for configuration reload

maheshkelkar:
References #120

### DIFF
--- a/features/api/updating.feature
+++ b/features/api/updating.feature
@@ -1,0 +1,19 @@
+Feature: DeployDB UPDATE APIs
+
+  As a RESTful client or user
+  I should be able to update DeployDB configuration
+
+  @wip
+  Scenario: Reloading the config while system is idle will succeed
+
+    Given configuration directory is "config"
+    When I POST to "/configReload"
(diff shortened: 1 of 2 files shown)


maheshkelkar:
@rtyler @GiriDandu Could you please review this?

features/api/updating.feature
```diff
@@ -0,0 +1,19 @@
+Feature: DeployDB UPDATE APIs
+
+  As a RESTful client or user
+  I should be able to update DeployDB configuration
+
+  @wip
+  Scenario: Reloading the config while system is idle will succeed
+
+    Given configuration directory is "config"
```
GiriDandu: Can we have this /config/reload that way in the future we can have /config/model/reload or /config/model/list. May be its too far fetched idea

GiriDandu: Also you may want to move this under /config/updating.feature from /api/updating.feature

rtyler:
I think we need to find a way to get this onto the admin servlet, but the API is fine as far as I'm concerned right now

Issue #44, opened 2015-02-13
DeployDB should be able to reload configuration with an API call


rtyler:
In order to support a CLI tool which might `deploydb reload` we should implement a simple REST call which will force DeployDB to reload *all* configuration files on disk.

Not sure where in the URL space this should live, but it might make sense to add it to the Admin app that Dropwizard runs side-by-side with the app itself

maheshkelkar:
The reload should fail if deployDb is actively deploygin i.e. if there are any active flows/deployments

maheshkelkar:
@rtyler if we have multiple instances of deployDB app running, do we expect admin app to upload config and reload config on all the instances?

rtyler:
@maheshkelkar that's a good question. I'm curious what @polynomial and @mbbx6spp think on this issue.

maheshkelkar:
@rtyler If we store configuration models in DB, then admin doesn't have to update all the instances. 

For uploading the config, another alternative is to have admin pass the URL of config file locations with reload API. The deployDB can turn around and attempt to read config files from that (http://<host>/services/<files>); However, I am not sure if we can GET file(s) or iterate thr files from a URL

rtyler:
@maheshkelkar configuration in a DB is poor design choice (IMHO) and one that causes a lot of frustration for operating a service.

I think it might be perfectly fine to make configuration reloading host specific, and then when we get around to having a CLI or something like that, the configuration reloading support would require an explicit host option

maheshkelkar:
@rtyler is this something that chef can do? That is uploading new config files and triggering configuration reload on all the instances?

rtyler:
@maheshkelkar yep!

maheshkelkar:
@rtyler I assume chef will copy the config to "configDirectory" location and use REST API to trigger the configuration reload. The REST API would not have any input (effectively an empty POST to the path). 

The deployDB will load the config from "configDirectory" specified in its configuration.

rtyler:
@maheshkelkar assumption correct!

PR #122, opened 2015-03-27
Added configReload admin task to reload the config, if no deployments are in progress

maheshkelkar:
Also:
- Fixed Deployment hashCode function causing HibernateExpcetion.
- Fixed the enumerated value stored for flow's status in DB

References #44

### DIFF
--- a/features/api/config.feature
+++ b/features/api/config.feature
@@ -15,7 +15,9 @@ Feature: DeployDB config APIs
     """
 
 
-  Scenario: When system has deployments in progress, attempt to reload config should fail
+  Scenario: If deployments are NOT verified i.e. are in progress, then attempt to reload
+            config should fail. A deployment is considered as in progress, if the status
+            is NOT SUCCESS or FAILED
 
     Given there is a deployment
     When I POST to "/tasks/configReload" from the admin app
--- a/features/api/config.feature
+++ b/features/api/config.feature
@@ -3,17 +3,25 @@ Feature: DeployDB config APIs
   As a DeployDB administrator
   I should be able to update DeployDB configuration
 
-  @wip
+
   Scenario: When the system is idle, reloading the shall succeed
 
-    Given configuration directory is "config"
+    Given DeployDb configuration directory path is "./config"
     When I POST to "/tasks/configReload" from the admin app
-    Then the response should be 204
+    Then the response should be 200
+    And the response body should be:
+    """
+    Done!
+    """
 
-  @wip
+
   Scenario: When system has deployments in progress, attempt to reload config should fail
 
     Given there is a deployment
     When I POST to "/tasks/configReload" from the admin app
-    Then the response should be 400
+    Then the response should be 200
+    And the response body should be:
+    """
+    Failed: Configuration reload is not allowed while deployments are in progress
+    """
 
--- a/src/main/groovy/deploydb/ConfigReloadTask.groovy
+++ b/src/main/groovy/deploydb/ConfigReloadTask.groovy
@@ -1,0 +1,50 @@
+package deploydb
+
+import com.codahale.metrics.annotation.Timed
+import com.google.common.collect.ImmutableMultimap
+import io.dropwizard.servlets.tasks.Task
+import org.hibernate.Session
+import org.hibernate.Transaction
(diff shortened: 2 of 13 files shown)


features/api/config.feature
```diff
@@ -3,17 +3,25 @@ Feature: DeployDB config APIs
...
-    Then the response should be 204
+    Then the response should be 200
+    And the response body should be:
+    """
+    Done!
+    """
+
 
-  @wip
   Scenario: When system has deployments in progress, attempt to reload config should fail
```
rtyler: Can you expand on why this is the case? Also does this mean every deployment must be either COMPLETED or CREATED in order for config reload to work?

src/main/groovy/deploydb/ConfigReloadTask.groovy
```diff
@@ -0,0 +1,50 @@
...
+    ConfigReloadTask(WorkFlow workFlow) {
+        super('configReload')
+        this.workFlow = workFlow
+    }
+
+    @Timed
+    @Override
+    void execute(ImmutableMultimap<String, String> parameters, PrintWriter output) throws Exception {
+
+        Session session = this.workFlow.deployDBApp.getSessionFactory().openSession()
```
rtyler: I don't understand the use of Session or Transactions below, we're not making any database related queries in this process (as far as I can tell/hope), so what's the transaction doing?

src/main/groovy/deploydb/ConfigReloadTask.groovy
```diff
@@ -0,0 +1,50 @@
...
+        try {
+            ManagedSessionContext.bind(session)
+            Transaction transaction = session.beginTransaction()
+            try {
+                this.workFlow.loadConfigModels(true)
+                transaction.commit()
+                output.println("Done!")
+                output.flush()
```
rtyler: Does output need to be flushed? I would *hope* that the caller of execute() would do that for us

rtyler:
> Fixed Deployment hashCode function causing HibernateExpcetion.

In the future, please send notable changes like that in a separate pull request.

features/api/config.feature
maheshkelkar: BTW, this is no different than commit in #120 

Yes. If every flow (i.e. deployment) is either in SUCCESS or FAILED state. (i.e.  until deployment is verified)

We are mainly protecting config change to affect the verification. We do not want deployment to complete but fail to verify due to config change

src/main/groovy/deploydb/ConfigReloadTask.groovy
```diff
@@ -0,0 +1,50 @@
...
+
+
+/**
+ * Admin task to reload DeployDB configuration
+ */
+class ConfigReloadTask extends Task {
+    private WorkFlow workFlow
+    private static final Logger logger = LoggerFactory.getLogger(ConfigReloadTask.class)
+
+    ConfigReloadTask(WorkFlow workFlow) {
```
maheshkelkar: done. will push once everything approved

src/main/groovy/deploydb/ConfigReloadTask.groovy
maheshkelkar: We cannot UnitOfWork on admin tasks.

We need to check DB for pending deployments/flows

src/main/groovy/deploydb/ConfigReloadTask.groovy
maheshkelkar: TaskServlet does call output.close(), I am not sure if that calls output.flush(). BTW all the other examples seem to flush(), hence I duplicated it

features/api/config.feature
maheshkelkar: How about this:

  Scenario: If deployments are NOT verified i.e. are in progress, then attempt to reload
            config should fail. A deployment is considered as in progress, if the status is NOT in SUCCESS or FAILED state

rtyler:
Bouncing back to Mahesh since there are some outstanding items I don't see addressed yet

rtyler:
I'm going to merge this now, but per some of our IRL discussions @maheshkelkar is going to scope out a change in the design for how we manage the relationship between configuration and a Flow

PR #132, opened 2015-03-31
Issue 129 generic query to fectch artifacts

GiriDandu:


src/main/groovy/deploydb/dao/ArtifactDAO.groovy
```diff
@@ -45,4 +45,27 @@ class ArtifactDAO extends AbstractDAO<Artifact> {
         }
         return null
     }
+
+    /**
+     * Locate an Artifact based on generic query parameter
+     *
+     * @param group A valid group name (e.g. "com.example")
+     * @param name The artifact's name (e.g. "dropwizard-core")
```
GiriDandu: cut paster error, params are wrong

src/main/groovy/deploydb/resources/ArtifactResource.groovy
```diff
@@ -48,6 +48,23 @@ public class ArtifactResource {
     }
 
     @GET
+    @UnitOfWork
+    @Timed(name = "get-requests")
+    List<Artifact> byName(@QueryParam("query") String type,
+                          @QueryParam("pageNumber") @DefaultValue("0") IntParam artifactPageNumber,
+                          @QueryParam("perPageSize") @DefaultValue("5") deploydb.ModelPageSizeParam
```
maheshkelkar: I think default value for pagesize should be 20. The value of 5 is very low IMO

src/main/groovy/deploydb/dao/ArtifactDAO.groovy
```diff
@@ -45,4 +45,27 @@ class ArtifactDAO extends AbstractDAO<Artifact> {
...
+
+        // logically "or" all the columns you want to search on
+        List<Artifact> artifacts = criteria().add( Restrictions.disjunction()
+                                               .add(Restrictions.like('name', type))
+                                               .add(Restrictions.like('group', type))
+                                               .add(Restrictions.like('version', type)))
+                                             .addOrder(Order.desc('createdAt'))
+                                             .setFirstResult(pageNumber)
+                                             .setMaxResults(perPageSize)
+                                             .list()
```
maheshkelkar: SHould we add a unique() here: to cover a scenarios where it finds a match for a string in name, as well as group for same artifact

features/api/artifact/reading.feature
```diff
@@ -165,7 +165,7 @@ Feature: Artifact READ APIs
       }
     """
 
-  @wip @freezetime
+  @query @freezetime
```
maheshkelkar: Perhaps I missed this exchange, but what is the purpose of query tag?

src/main/groovy/deploydb/resources/ArtifactResource.groovy
GiriDandu: ok, i will change it.

src/main/groovy/deploydb/dao/ArtifactDAO.groovy
GiriDandu: We don't need to make it unique(). The disjunction will make it unique. I verified using test example

features/api/artifact/reading.feature
GiriDandu: it's easy to test all "query" related test cases using this tag. Something similar to webhooks tag. If you think it is too much granular. I will remove it

maheshkelkar:
:+1:

PR #133, opened 2015-03-31
Add API to fetch deployment by artifact id

### LINKED ISSUE #117: API needed to access Deployments based on an Artifact ID
In order to present the "Developer View" we've discussed in chat (mockups still coming!), we'll need an API which, given an artifact ID, would return the deployments for that single artifact ID.

I'm inclined to suggest: `GET /api/deployments/by-artifact/{{art_id}}` unless there's a better suggestion.

GiriDandu:
- Fixes #117

### DIFF
--- a/features/api/deployment/reading.feature
+++ b/features/api/deployment/reading.feature
@@ -250,3 +250,42 @@ Feature: Deployment READ APIs
     Given there are deployments
     When I GET "/api/deployments?pageNumber=0&perPageSize=0"
     Then the response should be 400
+
+
+  @wip @freezetime
+  Scenario: Fetching deployment by artifact id
+
+    Given there is a deployment
+    When I GET "/api/deployments/by-artifact/1"
+    Then the response should be 200
+    And the body should be JSON:
+    """
+      [{
+        "id" : 1,
+        "artifact" : {
+          "id" : 1,
+          "group" : "com.example.cucumber",
+          "name" : "cucumber-artifact",
+          "version" : "1.0.1",
+          "sourceUrl" : "http://example.com/maven/com.example.cucumber/cucumber-artifact/1.0.1/cucumber-artifact-1.0.1.jar",
+          "createdAt" : "{{created_timestamp}}"
+        },
+        "environment" : "pre-prod",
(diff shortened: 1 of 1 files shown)


features/api/deployment/reading.feature
```diff
@@ -250,3 +250,42 @@ Feature: Deployment READ APIs
...
+          "promotion" : "jenkins-smoke",
+          "status" : "STARTED",
+          "infoUrl" : null,
+          "createdAt" : "{{created_timestamp}}"
+         }],
+        "createdAt" : "{{created_timestamp}}"
+      }]
+    """
+
+  @wip @error @freezetime
```
maheshkelkar: freezetime not needed. I dont think its invalid-d is passing string instead of integer. THis is a valid case, so we should omit error tag

features/api/deployment/reading.feature
```diff
@@ -250,3 +250,42 @@ Feature: Deployment READ APIs
     Given there are deployments
     When I GET "/api/deployments?pageNumber=0&perPageSize=0"
     Then the response should be 400
+
+
+  @wip @freezetime
+  Scenario: Fetching deployment by artifact id
+
+    Given there is a deployment
+    When I GET "/api/deployments/by-artifact/1"
```
maheshkelkar: Are we planning to add pageSize & pageNumber here? If we are planning, I think we can get it done as a part of this implementation.

features/api/deployment/reading.feature
GiriDandu: Agreed, i will remove it. It's CAP mistake

features/api/deployment/reading.feature
GiriDandu: I am not too sure. 
For a given artifact we will create as many deployments as environments for a given service that is artifact part of.  Generally we don't expect > 10 environments for all the services that given artifact is part of. So pagination support may not be required. 
But if we think we will create multiples of 10s of deployments for a given artifact, then I will add pagination support

maheshkelkar: ok

PR #135, opened 2015-04-01
Added docs for DeployDB launch and models configuration

maheshkelkar:


maheshkelkar:
@GiriDandu please review as well. I couldn't add you to assignee or notify list

GiriDandu:
>Models-Configuration Management: 
Can we have following sections. It's kind of easier to understand.
Goals: State the requirements. Operator is allowed to change the config anytime in the deployment flow etc
Design: How it is done - checksum etc
Limitations: Only allow to be backward compatible one version etc

maheshkelkar:
> Goals: State the requirements. Operator is allowed to change the config 
> anytime in the deployment flow etc
> Design: How it is done - checksum etc

ok, we can do that. But, is there a problem with current writeup. Is it falling short of communicating the intent.

> Limitations: Only allow to be backward compatible one version etc

Sorry, I am not sure I understand this comment. Can you please elaborate?

rtyler:
:+1:  from my side

Issue #126, opened 2015-03-30
Make deployments self sufficient by storing config information in the DB objects


maheshkelkar:
Once the deployment is created, it reads entire applicable configuration and thus making it impervious to any future changes to the configuration.

Common changes:
* We already absorb most of the config information in to DB models (i.e. flow, deployment & artifact)
* As a part of this issue, we will have to store webhooks & some minor imp information from other config models (i.e. Service, Environment, Pipeline, promotion, etc) into DB models (i.e. flow, deployment & artifact).
* To keep track of changes to config, we need to add version numbers. 
  * Option-1:
    * Internally or externally get 1 config-version representing any change 
    * Limited or no impact on REST APIs

  * Option-2:
    * A separate version for each config model change (i.e. service, webhook, pipeline, environemnt, promotion, etc)
    * Will have impact on all REST APIs

maheshkelkar:
This will have direct impact on #44

polynomial:
Per discussion with @maheshkelkar, should have a version per config, both as a field in the file and as a field on the objects returned.

rtyler:
@maheshkelkar for option 1:

> Internally or externally get 1 config-version representing any change 

I don't understand what this means, can you expand more? Is this just storing a copy of each version of configs and then referring to that globally?

and for option 2:

> Will have impact on all REST APIs

I don't understand how this will have an impact on the REST APIs at all, can you elaborate/quantify the impact a bit more?

maheshkelkar:
> I don't understand what this means, can you expand more? Is this just storing a 
> copy of each version of configs and then referring to that globally?
Top level config.yml has a "version" field and chef must bump it up for each change to underlying service, environemtn, etc

> I don't understand how this will have an impact on the REST APIs at all, 
> can you elaborate/quantify the impact a bit more?

*options-2 talks about adding a version field for each model (service, pipeline, promotion, environment, etc).
* E.g. Flow will have (artifact, serviceIdent, service version, pipelineIdent, pipelineIdent, deployments)
* E.g. Deployment will have (artifact, environemntIdent, environment version, webhooks)
- Hence, I need to update JSON representation of the REST APIs that return deployment, flow, environment, service, etc

PR #141, opened 2015-04-02
Allow Models-config to change (configReload) while deployments are in progress

maheshkelkar:
- Added ModelConfig model, DB table, etc to keep track of old configs.

- Added checksum calculations to keep track of each configReload
  iteration. This calculation reads all the directories and files
  in a deterministic order and computes checksum, which should remain
  same for given config on any platform.

- The checksum is stored in the DeployDB app object as well Flow object
  in order to associate them with the given iteration of config.

- On the subsequent Flow/Deployment transitions, if checksum from the
  Flow matches with that in the App object, then in-memory (i.e. registry)
  Model objects are used. Otherwise Model objects are read from the
  ModelConfig objects.

- In loadConfigModels, moved code repetitions such as loading from
  string, inserting into registry, checksum calculations, etc.
  to common function

- Added cucumber and spock tests to validate the code.

References #126


maheshkelkar:
@GiriDandu Please review.

features/api/config.feature
```diff
@@ -16,14 +16,143 @@ Feature: DeployDB config APIs
...
     Then the response should be 200
     And the response body should be:
     """
-    Failed: Configuration reload is not allowed while deployments are in progress
+    Done!
+    """
+
+
+  @freezetime
+  Scenario: If deployment is CREATED state and configuration is reloaded, then
```
rtyler: This scenario name is too long. Can this be summarized into less than 80 chars :p 

Test reporters will all truncate this after the first line which will make the reports not very useful

features/api/config.feature
```diff
@@ -16,14 +16,143 @@ Feature: DeployDB config APIs
...
+          "sourceUrl" : "http://example.com/maven/com.example.cucumber/cucumber-artifact/1.0.1/cucumber-artifact-1.0.1.jar",
+          "createdAt" : "{{created_timestamp}}"
+        },
+        "status" : "STARTED",
+        "service" : "faas",
+        "environment" : "pre-prod",
+        "createdAt" : "{{created_timestamp}}"
+      }
+    """
+    And the webhook 2 should be invoked with the JSON:
```
rtyler: I don't think there's any need to test the second webhook here right? We only need to verify that tihngs work the same for one of the webhooks that was configured right?

features/api/config.feature
```diff
@@ -16,14 +16,143 @@ Feature: DeployDB config APIs
...
+        "status" : "STARTED",
+        "service" : "faas",
+        "environment" : "pre-prod",
+        "createdAt" : "{{created_timestamp}}"
+      }
+    """
+
+
+  @freezetime
+  Scenario: If deployment is STARTED state and configuration is reloaded, then
```
rtyler: ETOOLONG

src/cucumber/groovy/deploydb/cucumber/StubAppRunner.groovy
```diff
@@ -90,12 +82,14 @@ public class StubAppRunner<C extends Configuration> {
...
+                            workFlow = application.workFlow
+
+                            /**
+                             * Setup config checksum for tests
                              */
-                            serviceRegistry = application.workFlow.serviceRegistry
-                            environmentRegistry = application.workFlow.environmentRegistry
-                            promotionRegistry = application.workFlow.promotionRegistry
-                            pipelineRegistry = application.workFlow.pipelineRegistry
+                            application.configChecksum = "0xdead"
```
GiriDandu: Not sure the reason for assignment.  Don't we calculate this and store it?

features/api/config.feature
maheshkelkar: ok. will do.

src/cucumber/groovy/step_definitions/DeploymentSteps.groovy
```diff
@@ -36,7 +36,7 @@ Given(~/^there is a deployment$/) { ->
         d1.addPromotionResult(p1)
 
         /* Create a flow */
-        Flow f = new Flow(a1, "faas")
+        Flow f = new Flow(a1, "faas", "0xdead")
```
GiriDandu: Is this because we don't have config in the DeploymentSteps?

features/api/config.feature
maheshkelkar: Yes. We don't. After adding one, I added another just to be sure. I can take that out.

features/api/config.feature
maheshkelkar: fixing

src/cucumber/groovy/deploydb/cucumber/StubAppRunner.groovy
maheshkelkar: For most of the cucumber tests, we do not load config, because it affects our test scenarios. So hardcoding it here, so that other checks performed on it don't fail

src/cucumber/groovy/step_definitions/WebhookSteps.groovy
```diff
@@ -5,28 +5,48 @@ import com.fasterxml.jackson.core.JsonToken
...
+                workFlow.deployDBApp.configChecksum, configBody,
+                workFlow.defaultIdent, ModelType.WEBHOOK)
+
+        withSession {
+            ModelConfigDAO modelConfigDAO = new ModelConfigDAO(sessionFactory)
+            modelConfigDAO.persist(modelConfig)
+        }
+    }
+
+    withTestWebhookServer { RequestWebhookObject requestWebhookObject ->
```
GiriDandu: Not too sure why this change is done.  Java docs to explain it?

src/cucumber/groovy/step_definitions/WebhookSteps.groovy
```diff
@@ -34,18 +54,11 @@ Given(~/^a (.*?) webhook "(.*?)" configuration:$/) { String webhookType,
...
-        ModelLoader<Webhook> webhookLoader = new ModelLoader<>(Webhook.class)
-        webhookManager.webhook = webhookLoader.loadFromString(configBody)
-
-        /*
          * Set the content type from the webhook and the event type. The content type will be
          * checked when deploydb invokes webhooks
          */
         requestWebhookObject.contentTypeParam = "application/vnd.deploydb."+
-                                                 webhookType+eventType+".v1+json"
+                webhookType+eventType+".v1+json"
```
GiriDandu: Not a big fan of this indenting. Also can you stick the code changes that you state in PR ;-) Open a new PR for code cleanup or say that some code cleanup is also in this PR.

src/cucumber/groovy/step_definitions/WebhookSteps.groovy
```diff
@@ -5,28 +5,48 @@ import com.fasterxml.jackson.core.JsonToken
...
 import webhookTestServer.models.RequestWebhookObject
 
+
 Given(~/^a (.*?) webhook "(.*?)" configuration:$/) { String webhookType,
                                                      String eventType, String configBody ->
 
     List<String> paths = getUrlPathFromWebhookConfigBody(configBody, eventType)
+    withWorkFlow { WorkFlow workFlow ->
+        /*
+         * Load the webhook configuration in webhookManager
```
GiriDandu: Fix the comment because you don't have webhookManager anymore

src/main/groovy/db/migration/V8__create_configs_table.groovy
```diff
@@ -0,0 +1,53 @@
...
+        /* Sql commands */
+        List<String> commands = []
+
+        /*
+         * Add modelConfigs table
+         */
+        commands += """
+            CREATE TABLE modelConfigs (
+                id BIGINT AUTO_INCREMENT,
+                checksum TEXT NOT NULL,
```
GiriDandu: Why not make checksum VARCHAR(512) instead of TEXT?

src/main/groovy/db/migration/V8__create_configs_table.groovy
```diff
@@ -0,0 +1,53 @@
...
+                deletedAt TIMESTAMP NULL,
+                PRIMARY KEY (id)
+            );
+        """
+
+        /*
+         * Add checksum column to flows table
+         */
+        commands += """
+            ALTER TABLE flows ADD COLUMN checksum TEXT;
```
GiriDandu: If you do change checksum data type in modelConfigs table, make sure you change it flows table also

src/main/groovy/deploydb/ConfigReloadTask.groovy
```diff
@@ -31,23 +28,15 @@ class ConfigReloadTask extends Task {
...
-        Session session = this.workFlow.deployDBApp.getSessionFactory().openSession()
-        try {
-            ManagedSessionContext.bind(session)
-            Transaction transaction = session.beginTransaction()
+        this.workFlow.deployDBApp.withHibernateSession() {
             try {
-                this.workFlow.loadConfigModels(true)
-                transaction.commit()
+                this.workFlow.loadConfigModels()
                 output.println("Done!")
```
GiriDandu: can you use standard logging instead of stdout. Also "Done" is not very descriptive for logging

src/main/groovy/deploydb/ConfigReloadTask.groovy
```diff
@@ -31,23 +28,15 @@ class ConfigReloadTask extends Task {
...
+            } catch (Exception e) {
+                logger.error("failed to reload the config with an exception: ", e)
+                output.println("Failed: " + e.getMessage())
```
GiriDandu: Remove this output.println()

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -59,21 +64,58 @@ public class WorkFlow {
...
+            filenames.each() { String filename ->
+                File modelFile = new File(modelDirectory, filename)
                 try {
-                    c.call(modelFile)
+                    /** Read YAML file into model object */
+                    def model = loader.load(modelFile)
+
+                    /** If registry is available, then insert model object into registry */
+                    String ident = defaultIdent
+                    if (registry) {
```
GiriDandu: When will the registry not present in this function?

src/cucumber/groovy/step_definitions/DeploymentSteps.groovy
maheshkelkar: Yes

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -189,31 +221,53 @@ public class WorkFlow {
...
 
-                tmpWebhook = this.webhookLoader.load(modelFile)
-                logger.debug("Loaded webhooks model from: ${modelFile.name}")
+                logger.debug("Loaded webhook model")
 
-                /* Now that we have found a valid webhook, we are done */
+                /* Now that we have found a valid global webhook, we are done */
                 throw new BreakLoopException()
             }
         } catch (BreakLoopException e) {
```
GiriDandu: Why there is no catch (all) here?

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -189,31 +221,53 @@ public class WorkFlow {
...
+        }
+        if (this.deployDBApp.configChecksum == newConfigChecksum) {
+            logger.info("Ignoring as no change in DeployDB Model Configuration detected")
+            return
+        }
+        this.deployDBApp.configChecksum = newConfigChecksum
+
+        /* At least one service MUST be configured for deployDb to function properly */
+        if (tmpServiceRegistry.getAll().isEmpty()) {
+            logger.info("NO SERVICES ARE CONFIGURED. DeployDB would not function properly")
```
GiriDandu: I think we should move this log category to at least WARN or higher.  We want to get the operator attention here

src/cucumber/groovy/step_definitions/WebhookSteps.groovy
maheshkelkar: * webhook object reference is moved from WebhookManager to WorkFlow, hence we dont need it
* This particular cucumber "step" is used in used in a scenarios-3 in config.feature (create flow, reload config, trigger deployment STARTED). Hence, we are storing the webhook and environment in the ModelConfig

src/main/groovy/deploydb/WorkFlow.groovy
```diff
@@ -221,7 +275,72 @@ public class WorkFlow {
...
+         * rebuild objects from DB
+         */
+        if (deployment.flow.configChecksum != this.deployDBApp.configChecksum) {
+            /* Load webhook from config */
+            models.ModelConfig webhookConfig =
+                    this.modelConfigDAO.findModelConfig(ModelType.WEBHOOK,
+                            defaultIdent, deployment.flow.configChecksum)
+            if (webhookConfig) {
+                models.Webhook.Webhook fetchedWebhook =
+                        this.webhookLoader.loadFromString(webhookConfig.contents)
```
GiriDandu: Are we certain that this won't throw an exception?

src/cucumber/groovy/step_definitions/WebhookSteps.groovy
maheshkelkar: * This is Intellij's doing
* If I had done it, I am not going to put these in-consequential but appropriate changes stated in the PR ;-)

src/cucumber/groovy/step_definitions/WebhookSteps.groovy
maheshkelkar: ok

src/main/groovy/db/migration/V8__create_configs_table.groovy
maheshkelkar: I agree this needs to be changed. There is inconsistency in use of VARCHAR(8192) or TEXT everywhere in the code. If I had used VARCHAR, why not TEXT, was a valid question to. 

Let me investigate if there is a limit on output of md5hex() output, if yes, I will put it to that size OR just change it to VARCHAR

src/main/groovy/db/migration/V8__create_configs_table.groovy
maheshkelkar: ok

src/main/groovy/deploydb/ConfigReloadTask.groovy
maheshkelkar: This "output" is return field for admin "Task". This was reviewed as a part of #120. If it needs to be addressed, we can evaluate it later

src/main/groovy/deploydb/ConfigReloadTask.groovy
maheshkelkar: We can't, this is a response back to the "ConfigReload" task

src/main/groovy/deploydb/WorkFlow.groovy
maheshkelkar: In case of webhook, the registry is not passed, because its singular. In that case "defaultIdent" is used.

src/main/groovy/deploydb/WorkFlow.groovy
maheshkelkar: This was a part of #44. @rtyler wanted it to be handled by the caller.

src/main/groovy/deploydb/WorkFlow.groovy
maheshkelkar: This code is unchanged (simply moved around due to new checksum calculations). But, WARN is too strong to do so.

src/main/groovy/deploydb/WorkFlow.groovy
maheshkelkar: I don't expect any known exceptions. Similar to other processing, it will abort the process and handled by the caller (i.e. resource in most of the cases)

maheshkelkar:
Hi @mbbx6spp, could you please review the schema changes in this commit

rtyler:
Looks fine to me but needs to be synced with master looks like

GiriDandu:
:+1:

PR #149, opened 2015-04-06
Issue 127 travis testing using multiple database providers

GiriDandu:


deploydb.mysql.cucumber.yml
```diff
@@ -0,0 +1,25 @@
+database:
```
maheshkelkar: Can we move all there cucumber yml files under src/tests/resources/cucumber/ OR perhaps club them under a common directory (like we did in case of example)?

Notet that I am going to add another cucumber file for ldap, soon.

features/api/artifact/reading.feature
```diff
@@ -60,7 +60,7 @@ Feature: Artifact READ APIs
     Then the response should be 400
 
 
-  @freezetime
+  @freezetime 
   Scenario: Fetching an artifact by the (group, name) tuple
```
maheshkelkar: trailing whitespace

features/api/artifact/reading.feature
```diff
@@ -78,7 +78,7 @@ Feature: Artifact READ APIs
       }]
     """
 
-  @freezetime
+  @freezetime
```
maheshkelkar: trailing whitespace

deploydb.mysql.cucumber.yml
GiriDandu: Why don't you like the files in the top level directory ;-) You always like to hide them in a sub directory.

 They are easy to find in the top level directory

src/main/groovy/deploydb/models/AbstractModel.groovy
```diff
@@ -16,7 +16,7 @@ abstract class AbstractModel {
     Long id
 
     @Column(name='createdAt', nullable=false)
-    protected DateTime createdAt = DateTime.now()
+    protected DateTime createdAt = DateTime.now().withMillisOfSecond(0)
```
maheshkelkar: I agree that this is needed for cucumber. But, now we are forcing milliseconds to be "0" for runtime as well. Right? Anyway we can avoid it?

GiriDandu: Nope you can't avoid it. Given all the workaround you need to take care of timestamp precision in travis, this can't be done unless travis moves to mysql 5.6.4 for default configuration

deploydb.mysql.cucumber.yml
maheshkelkar: It looks un-organized :)

maheshkelkar:
:+1:

PR #150, opened 2015-04-07
Upgrade to Dropwizard 0.8.1

rtyler:


maheshkelkar:
@GiriDandu If logback issue is fixed, should we backout workaround made in 0.8.0 commit

GiriDandu:
@maheshkelkar cucumber archiving doesn't make sense. That change should not be removed.

PR #151, opened 2015-04-07
Fix Gradle JDBC driver dependencies to be runtime

mbbx6spp:
It is generally bad practice to define JDBC or other dynamically loaded with
standard API to be required as compile time dependencies as the purpose of
these is to load them in at runtime and only rely on their generic API. When
you define a dependency like this as compile scope instead of runtime you
open yourself up to developers using MySQL, or H2 specific APIs without
realizing this dependency because the build will never catch it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,9 +50,9 @@ dependencies {
     /* Webhook support */
     compile 'com.github.lookout:whoas:0.2.0'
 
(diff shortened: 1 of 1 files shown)


rtyler:
Good change, i'm looking forward to removing the MySQL runtime dependency and requiring users who wish to use any JDBC connector to provide the classes on the classpath though :)

Issue #162, opened 2015-04-21
Implement manual LDAP promotion functionality


maheshkelkar:
References #159

PR #165, opened 2015-04-22
Added implementation for manual LDAP promotion

maheshkelkar:
- Added PromotionImpl interface for promotion specific functions 
- Added derived ManualLDAPPromotionImpl with User authorization
- Removed "Auth" from GET deployments APIs (it was added as an experiment)
- Added "Auth" for POST promotion results

References #162


maheshkelkar:
@GiriDandu please review

src/main/groovy/deploydb/models/Promotion/Promotion.groovy
```diff
@@ -0,0 +1,126 @@
...
+     * InstantiationException - if this Class represents an abstract class, an interface,
+     *                          an array class, a primitive type, or void
+     *                          or if the class has no nullary constructor
+     *                          or if the instantiation fails for some other reason.
+     */
+    private PromotionImpl getPromotionImpl() throws
+            ClassNotFoundException,
+            InstantiationException,
+            IllegalAccessException {
+        return ClassLoader.getSystemClassLoader().loadClass(this.type).getDeclaredConstructor().newInstance()
```
rtyler: From my read on this, this will lazily evaluate at runtime, meaning that configuration errors will cause an exception when a manual promotion is created as opposed to instantiation of this object.

IMO this is undesirable behavior and will probably annoy users :P

rtyler:
Will you make sure the `Promotion` directory is lower cased? all directories should be lowercased, as should the package names (class names have the capitalization)

src/main/groovy/deploydb/models/Promotion/Promotion.groovy
maheshkelkar: We can validate the class name during the initialization (i.e. while reading model config) and reject the promotion if it does not have a valid "type"

maheshkelkar: ok, I was able to fix this using: ValidationMethod. This throws a ConfigurationValidationException.

maheshkelkar:
@rtyler @GiriDandu I have taken care of the comments, I have merged up with Giri's changes. Let me know if you have any other pending comments, or I will merge it when Travis goes green.

GiriDandu:
No more comments from my side

src/main/groovy/deploydb/models/promotion/Promotion.groovy
```diff
@@ -0,0 +1,141 @@
...
+    @JsonProperty
+    String type
+
+    /**
+     * Validate contents of "type" using following annotation
+     *
+     * @return true if its a valid class name
+     */
+    @ValidationMethod(message = "Promotion does not have a valid class name in \"type\"")
+    boolean isType() {
```
rtyler: It's not clear to me from the changes here, where is `isType()` called such that the error will be displayed at load time instead of runtime?

maheshkelkar: @rtyler 
* The error is displayed at the load time (simple stderr; it raises an exception (see below) and our loadConfigFile will document that immediately)
* @ValidationMethod is a dropwizard annotation. 
* The format for the method name and signature is fixed: 
boolean "is<variable name>"()
* This method is called along with other validation annotations, such as @NotEmpty, @Size, etc
* Currently, we are hitting this when parsing yaml into object (raising ConfigurationValidationException)
* I have added a test for this (first)
* Here are couple of links to read more about it:
http://dropwizard.github.io/dropwizard/0.6.2/manual/core.html#advanced

rtyler: @maheshkelkar will you update the isType() javadoc and indicate that this is for dropwizard's validations

maheshkelkar: ok will do.

maheshkelkar:
@rtyler done. I have added the javadocs. Let me know if you want to have another pass, or I will merge it tomorrow if Travis is green.

Issue #166, opened 2015-04-22
Enable LDAP configuration in all cucumber specific deploydb config files


maheshkelkar:
If LdapAuthentication is configured, authorization fails for ManualLDAPPromotion (because User is null, and allowedGroup is configured). As a result cucumber test fails. 

Investigate into a cucumber tag that will allow us to make that scenario optional

maheshkelkar:
The main problem is we have a scenario and that fails if authentication is not configured. And that is correct behavior. But, cucumber does not allow us to say if enabled check this or check that. 

There is no mechanism available in Cucumber JVM to conditionally skip a scenanrio. You can throw PendingExpcetion in middle of test, that skips the scenario, but shows up all "red" in the test and does not look right. 

I discussed with @GiriDandu and we decieded that instead we will add Ldap configuration to all flavors of deploydb config files (i.e. h2, mysql & postgress specifics) used for cucumber

PR #173, opened 2015-05-03
Add ldap config to all deploydb config files used in cucumber tests

maheshkelkar:
Also, make the manual LDAP authentication scenario self-explanatory.

References #166


maheshkelkar:
@rtyler please review

GiriDandu:
:+1:

Issue #172, opened 2015-05-01
deployment is prematurely marked successful on the first of many promotionResult add


maheshkelkar:
If a deployment has multiple promotions and first promotion results are added (w/ success) deployment is marked as success in error. In reality, it should wait until results of all the promotions are added successfully 

```
{
    "artifact": {
        "createdAt": "2015-05-01T14:17:50.000Z",
        "group": "adv.group.1",
        "id": 1,
        "name": "bg1",
        "sourceUrl": "http://example.com/cucumber1.jar",
        "version": "2.3.4.5"
    },
    "createdAt": "2015-05-01T14:17:50.000Z",
    "deployments": [
        {
            "artifact": {
                "createdAt": "2015-05-01T14:17:50.000Z",
                "group": "adv.group.1",
                "id": 1,
                "name": "bg1",
                "sourceUrl": "http://example.com/cucumber1.jar",
                "version": "2.3.4.5"
            },
            "createdAt": "2015-05-01T14:17:50.000Z",
            "environment": "basicEnv",
            "id": 1,
            "promotions": [
                {
                    "createdAt": "2015-05-01T14:17:50.000Z",
                    "id": 1,
                    "infoUrl": null,
                    "promotion": "manualPromo",
                    "status": "STARTED"   <<<<<<< PENDING
                },
                {
                    "createdAt": "2015-05-01T14:17:50.000Z",
                    "id": 2,
                    "infoUrl": null,
                    "promotion": "basicPromo",
                    "status": "SUCCESS"
                }
            ],
            "service": "advServ",
            "status": "SUCCESS"  <<<<<<<<<< ALREADY MARKED AS SUCCESS
        }
    ],
    "id": 1,
    "service": "advServ"
}
```

PR #174, opened 2015-05-04
Fix a case where deployment with multiple promotions is prematurely marked as SUCCESS

maheshkelkar:
References #172

### DIFF
--- a/src/test/groovy/deploydb/PromotionCompletedNotificationSpec.groovy
+++ b/src/test/groovy/deploydb/PromotionCompletedNotificationSpec.groovy
@@ -215,4 +215,45 @@ class PromotionCompletedNotificationsSpec extends Specification {
         sleep(1000)
         integAppHelper.webhookRunner.requestWebhookObject.requestMessageBodies.size() == 4
     }
+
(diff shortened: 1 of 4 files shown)


maheshkelkar:
@rtyler I have assigned it to Giri. But, please take a look.

src/test/groovy/deploydb/PromotionCompletedNotificationSpec.groovy
```diff
@@ -215,4 +215,44 @@ class PromotionCompletedNotificationsSpec extends Specification {
...
+        Long deploymentId = 1
+        integAppHelper.withSession {
+            deployment = integAppHelper.runner.getApplication().workFlow.deploymentDAO
+                    .get(deploymentId)
+        }
+
+        then:
+        success == true
+        sleep(1000)
+        deployment != null
```
GiriDandu: This check is not required. If the deployment doesn't exist, then the sendPromotionCompletedTrigger() should fail.

maheshkelkar: Yeah. This check is only ensuring that deploymentDAO.get(deploymentId) has worked.

PR #185, opened 2015-06-02
Using FetchMode.SUBSELECT annotation on OneToMany relationship forces hibernate to not use the Join.

### LINKED ISSUE #183: Deployments are rendered twice
This is the output of api/deployments:
https://gist.githubusercontent.com/emilong/fb43b7e5862ba159eb5c/raw/53e438c44456841bef797622e0a1c7a73187fe24/gistfile1.json
Notice that models with the same id appear twice.

Steps to repro:
- get this branch: https://github.com/emilong/deploydb/tree/multi-env-pipeline
- ./gradlew shadowJar
- java -jar build/libs/deploydb-*-all.jar db migrate complicated/deploydb.launch.yml
- java -jar build/libs/deploydb-*-all.jar server complicated/deploydb.launch.yml
- post this:
```shell
curl -X POST -H  "Content-Type: application/json" http://localhost:8080/api/artifacts -d "{
  \"group\" : \"immortan.joe\",
  \"name\": \"mask\",
  \"version\" : \"da39a3ee5e6b4b0d3255bfef95601890afd80709\",
  \"sourceUrl\" : \"http://example.com/immortan.jar\"
}"
```
- curl http://localhost:8080/api/deployments

maheshkelkar:
This FetchMode instead sends multiple sql calls to the db to retrieve the data; thus eliminates duplicates.

Fixes #183

### DIFF
--- a/src/main/groovy/deploydb/dao/ArtifactDAO.groovy
+++ b/src/main/groovy/deploydb/dao/ArtifactDAO.groovy
@@ -39,7 +39,6 @@ class ArtifactDAO extends AbstractDAO<Artifact> {
 
         List<Artifact> artifacts = criteria().add(Restrictions.eq('group', group))
                           .add(Restrictions.eq('name', name))
-                          .addOrder(Order.desc('createdAt'))
                           .addOrder(Order.asc('id')).list()
 
         if(artifacts.size() > 0) {
--- a/src/main/groovy/deploydb/dao/ArtifactDAO.groovy
+++ b/src/main/groovy/deploydb/dao/ArtifactDAO.groovy
@@ -39,7 +39,9 @@ class ArtifactDAO extends AbstractDAO<Artifact> {
 
         List<Artifact> artifacts = criteria().add(Restrictions.eq('group', group))
                           .add(Restrictions.eq('name', name))
-                          .addOrder(Order.desc('createdAt')).list()
+                          .addOrder(Order.desc('createdAt'))
+                          .addOrder(Order.asc('id')).list()
+
         if(artifacts.size() > 0) {
             return artifacts.last()
         }
--- a/src/main/groovy/deploydb/dao/DeploymentDAO.groovy
+++ b/src/main/groovy/deploydb/dao/DeploymentDAO.groovy
@@ -37,7 +37,8 @@ class DeploymentDAO extends AbstractDAO<Deployment> {
     Deployment getLatest() {
 
         List<Deployment> deployments = criteria()
-                .addOrder(Order.desc('createdAt')).list()
+                .addOrder(Order.desc('createdAt'))
+                .addOrder(Order.asc('id')).list()
         if (deployments.size() > 0) {
             return deployments.last()
         }
--- a/src/main/groovy/deploydb/dao/DeploymentDAO.groovy
+++ b/src/main/groovy/deploydb/dao/DeploymentDAO.groovy
@@ -37,7 +37,6 @@ class DeploymentDAO extends AbstractDAO<Deployment> {
     Deployment getLatest() {
 
         List<Deployment> deployments = criteria()
-                .addOrder(Order.desc('createdAt'))
                 .addOrder(Order.asc('id')).list()
         if (deployments.size() > 0) {
             return deployments.last()
--- a/src/test/groovy/deploydb/ModelConfigHelper.groovy
+++ b/src/test/groovy/deploydb/ModelConfigHelper.groovy
@@ -218,7 +218,21 @@ promotions:
         createServiceConfigFile()
     }
 
+    def createMultiPromoServiceModelConfigFiles() {
+        createPromotionConfigFile()
+        createManualPromotionConfigFile()
+        createPipelineConfigFile()
+        createMultiPromoServiceConfigFile()
+    }
 
+    def createMultiPromoMultiEnvServiceModelConfigFiles() {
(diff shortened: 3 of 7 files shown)


maheshkelkar:
@rtyler @emilong Please review this cucumber API definition.

src/main/groovy/deploydb/dao/ArtifactDAO.groovy
```diff
@@ -39,7 +39,9 @@ class ArtifactDAO extends AbstractDAO<Artifact> {
 
         List<Artifact> artifacts = criteria().add(Restrictions.eq('group', group))
                           .add(Restrictions.eq('name', name))
-                          .addOrder(Order.desc('createdAt')).list()
+                          .addOrder(Order.desc('createdAt'))
+                          .addOrder(Order.asc('id')).list()
```
GiriDandu: Not sure I understand why we need to change another order by clause is needed? To fetch the latest, we should just order by createdAt and pick the latest

BTW, why is this change made? How does the deployment duplicate affect this?

src/main/groovy/deploydb/dao/DeploymentDAO.groovy
```diff
@@ -37,7 +37,8 @@ class DeploymentDAO extends AbstractDAO<Deployment> {
     Deployment getLatest() {
 
         List<Deployment> deployments = criteria()
-                .addOrder(Order.desc('createdAt')).list()
```
GiriDandu: Same comment as before

src/main/groovy/deploydb/dao/ArtifactDAO.groovy
maheshkelkar: * what test failed - /api/deployments/latest 
* why failed - if you have 2 deployments, both of them have the same createdAt time, then "latest" * can be either "id=1" or "id=2"; caused test failed
* why did it fail now - this is bound to fail at some point. we added a new annotation FetchMode on Deployment (perhaps that could have an impact)
* how does this change fixes that - we have added 2 levels of sorting (first by "createdAt" time, second by "id")

src/main/groovy/deploydb/dao/DeploymentDAO.groovy
maheshkelkar: See above. Let me know if you have any questions

src/test/groovy/deploydb/ModelConfigHelper.groovy
```diff
@@ -218,7 +218,21 @@ promotions:
...
     }
 
+    def createMultiPromoServiceModelConfigFiles() {
+        createPromotionConfigFile()
+        createManualPromotionConfigFile()
+        createPipelineConfigFile()
+        createMultiPromoServiceConfigFile()
+    }
 
+    def createMultiPromoMultiEnvServiceModelConfigFiles() {
```
GiriDandu: Where is this function used?

src/main/groovy/deploydb/dao/ArtifactDAO.groovy
GiriDandu: @maheshkelkar i don't agree that when we create two deployments we have the same createdAt value.  BTW there is no precision in the "TIMESTAMP" in the DDL.

BTW, the change we made is between Deployment <--> PromotionResult one-to-many relationship. Why should that affect the fetching latest Artifact.

maheshkelkar: @GiriDandu Yes, that change broke it for postgress. 

I believe we create timestamp with "0" millisecond value. I think you made that change exclusively to make it work for postgres/mysql

src/test/groovy/deploydb/ModelConfigHelper.groovy
maheshkelkar: This was used for flowReadingSpec. Eventually I took that out, but I have already pushed it with another push request.

src/main/groovy/deploydb/dao/ArtifactDAO.groovy
maheshkelkar: Here is the link for failure in postgress:
https://travis-ci.org/lookout/deploydb/jobs/65115691

GiriDandu: Why don't we test with postgres DB type in local setup. I will help you with that setup. It's not difficult. I know the precision of the TIMESTAMP in postgres is 9.4 is 1 millisecond is this root cause of the problem?

GiriDandu: I looked at travis log, it shows failure for deployment latest, not artifact latest

src/test/groovy/deploydb/ModelConfigHelper.groovy
GiriDandu: :-)

src/main/groovy/deploydb/dao/ArtifactDAO.groovy
GiriDandu: Please look at the DDL in the db.migration directory, there is no TIMESTAMP precision in them.  The only change special for postgres is ID generation.

maheshkelkar: Lets talk friday. I will walk you through the thinking behind this.

PR #188, opened 2015-06-02
Added Resource and DAO code to handle REST API to get all flows

### LINKED ISSUE #187: Add REST API to fetch all the flows
This API will return all the flows in the system. This will also allow us to add tests and fix the bug in hibernate join fetch when using OneToMany associations between flow & deployments

maheshkelkar:
Fixes #187

### DIFF
--- a/src/main/groovy/deploydb/dao/FlowDAO.groovy
+++ b/src/main/groovy/deploydb/dao/FlowDAO.groovy
@@ -38,10 +38,6 @@ class FlowDAO extends AbstractDAO<Flow> {
                 .setFirstResult(pageNumber)
                 .setMaxResults(perPageSize)
                 .addOrder(Order.asc('id')).list()
-
-        for (Flow flow : flows) {
-            println "***MVK: flow = ${flow}"
-        }
         return flows
     }
 }
--- a/src/main/groovy/deploydb/dao/FlowDAO.groovy
+++ b/src/main/groovy/deploydb/dao/FlowDAO.groovy
@@ -4,6 +4,7 @@ import deploydb.models.Flow
 import groovy.transform.InheritConstructors
 import io.dropwizard.hibernate.AbstractDAO
 import org.hibernate.Session
+import org.hibernate.criterion.Order
 import org.hibernate.criterion.Restrictions
 
 /**
@@ -23,5 +24,24 @@ class FlowDAO extends AbstractDAO<Flow> {
 
         // now delete the found flow
         session.delete(flow)
+    }
+
+    /**
+     * Locate an Flow based on the (group, name) pair
+     *
+     * @param pageNumber A valid page number for pagination
+     * @param perPageSize A valid per page size
+     */
+    List<Flow> getByPage(int pageNumber, int perPageSize) {
+
+        List<Flow> flows = criteria()
+                .setFirstResult(pageNumber)
+                .setMaxResults(perPageSize)
+                .addOrder(Order.asc('id')).list()
+
+        for (Flow flow : flows) {
+            println "***MVK: flow = ${flow}"
+        }
+        return flows
     }
 }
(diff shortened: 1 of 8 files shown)


src/main/groovy/deploydb/dao/FlowDAO.groovy
```diff
@@ -24,4 +25,23 @@ class FlowDAO extends AbstractDAO<Flow> {
...
         session.delete(flow)
     }
+
+    /**
+     * Locate an Flow based on the (group, name) pair
+     *
+     * @param pageNumber A valid page number for pagination
+     * @param perPageSize A valid per page size
+     */
+    List<Flow> getByPage(int pageNumber, int perPageSize) {
```
GiriDandu: Can we change the name to getAll

maheshkelkar: This is consistent with deploymentDAO. And we cannot name get(), we already have a method for that.

maheshkelkar:
@rtyler Looks like Giri reviewed it, but he is on PTO now. Can you please take a quick look at it?

src/main/groovy/deploydb/dao/FlowDAO.groovy
```diff
@@ -24,4 +25,23 @@ class FlowDAO extends AbstractDAO<Flow> {
...
+
+        List<Flow> flows = criteria()
+                .setFirstResult(pageNumber)
+                .setMaxResults(perPageSize)
+                .addOrder(Order.asc('id')).list()
+
+        for (Flow flow : flows) {
+            println "***MVK: flow = ${flow}"
```
rtyler: println statements have no place in the codebase. This looks like forgotten debugging code (use logback!)

src/main/groovy/deploydb/models/Flow.groovy
```diff
@@ -29,9 +33,15 @@ class Flow extends AbstractModel {
     @JsonProperty
     Artifact artifact
 
+    /**
+     * Use of FetchMode.SUBSELECT annotation forces hibernate to not use the Join,
+     * and instead sends multiple sql calls to the db to retrieve the data,
+     * thus eliminating the duplicates.
+     */
```
rtyler: This comment is above the `@OneToMany` annotation which has `fetch=FetchType=EAGER`

I see the `@Fetch` annotation below that this comment seems to apply to. Not being too familiar with Hibernate, is the latter annotation overwriting the first one in `@OneToMany`?

src/test/groovy/deploydb/FlowReadingSpec.groovy
```diff
@@ -0,0 +1,60 @@
...
+                "application/vnd.deploydb.deploymentcreated.v1+json"
+    }
+
+    def cleanup() {
+        integAppHelper.stopApp()
+        integAppHelper.stopWebhookTestServerApp()
+        mcfgHelper.cleanup()
+    }
+
+    boolean verifyGetFlows() {
```
rtyler: This should have some Javadocs on what "verification" actually means (i.e. that it had a 200 and a single flow)

src/test/groovy/deploydb/FlowReadingSpec.groovy
```diff
@@ -0,0 +1,60 @@
...
+        // Create the required config
+        mcfgHelper.createMultiPromoMultiEnvServiceModelConfigFiles()
+
+        // load up the configuration
+        integAppHelper.runner.getApplication().loadModelConfiguration()
+
+        // create the artifact
+        integModelHelper.sendCreateArtifact()
+
+        when:
```
rtyler: Since this test case is just checking the truthyness of `verifyGetFlows()` you can shorten it FWIW:

```groovy
given:
/* things */

expect:
verifyGetFlows()
```

src/test/groovy/deploydb/FlowReadingSpec.groovy
rtyler: I'll also add, I think the function name does "read" quite right, since it's returning a boolean value, I think `areFlowsValid()` or `isFlowValid` make more sense

rtyler:
:-1:  right now, there's some fixes that need to go in

maheshkelkar:
@rtyler Thanks. I have taken care of the comments.

